PR #268, opened 2014-08-07
Excludes on stats

juergen-vogel:
Since Solr 4.8 you can add excludes to your stats fields in Solr. See https://issues.apache.org/jira/browse/SOLR-3177 for more information on these changes. 

I added the possibility to add excludes on a stats field in solarium for using this feature.


eostis:
Is it just me, or the pull request vanished from the code line ?

Too bad, because it solves a common issue with price filters with sliders (min-max filter): keeping both excluded (min, max) values, and filtered (from, to) values.

PR #273, opened 2014-09-12
Escape the phrases in a range query.

nubs:
In order to prevent query injection, the helper methods should escape
the phrases.  Otherwise you could end up with something like
`rangeQuery('1', '2] someOtherField:[* TO *')` and leak information, etc.

### DIFF
--- a/library/Solarium/Core/Query/Helper.php
+++ b/library/Solarium/Core/Query/Helper.php
@@ -216,6 +216,9 @@ class Helper
             $to = '*';
         }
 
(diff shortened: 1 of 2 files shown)


nubs:
Something like this should still probably be done in the newest code as well.  Leaving the library open to query injection could cause problems for folks.

thePanz:
This PR was not closed intentionally, sorry! (reason: we deleted the `develop` branch, and looks like GitHub closes all the PR against it).

Could you please reopen this PR against the latest `master` branch?

Issue #301, opened 2014-12-23
extract query error: the request was rejected because no multipart boundary was found


seanchen:
Hi,

I am using the [example code](https://github.com/solariumphp/solarium/blob/master/examples/2.7-extract-query.php) to test the Solarium Extract query on the following configuration:
- Solr 4.0.0
- PHP 5.6.2
- Solarium 3.3.0

I always got the following Exception on Tomcat / Solr:

```
SEVERE: null:org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found
        at org.apache.commons.fileupload.FileUploadBase$FileItemIteratorImpl.<init>(FileUploadBase.java:931)
        at org.apache.commons.fileupload.FileUploadBase.getItemIterator(FileUploadBase.java:331)
        at org.apache.commons.fileupload.FileUploadBase.parseRequest(FileUploadBase.java:349)
        at org.apache.commons.fileupload.servlet.ServletFileUpload.parseRequest(ServletFileUpload.java:126)
        at org.apache.solr.servlet.MultipartRequestParser.parseParamsAndFillStreams(SolrRequestParsers.java:344)
        at org.apache.solr.servlet.StandardRequestParser.parseParamsAndFillStreams(SolrRequestParsers.java:397)
        at org.apache.solr.servlet.SolrRequestParsers.parse(SolrRequestParsers.java:115)
        at org.apache.solr.servlet.SolrDispatchFilter.doFilter(SolrDispatchFilter.java:260)
        at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:241)
        at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:208)
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:220)
        at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:122)
        at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:170)
        at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:98)
        at org.apache.catalina.valves.AccessLogValve.invoke(AccessLogValve.java:950)
        at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:116)
        at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:408)
[...]

PR #303, opened 2014-12-23
some changes will debugging issue #301

seanchen:
some changes will I am debugging for issue #301.

- the CURL0PT_POSTFIELDS options should have the key => value format.
- adding test case to make sure the content-type header is set properly for Extract query.


basdenooijer:
Merged in PR #312 (because the target needed to be changed to develop branch), thanks for your contribution to Solarium!

PR #304, opened 2015-01-09
[WIP] Component refactoring

smoench:
I started this a while ago, because I was confused about placement and structure of those components.
The goal of this PR is to have all related classes of a component in its own folder.

It's not ready yet.


basdenooijer:
First of all thanks a lot for your effort, and sorry for not responding sooner!

I think the modeling of Solarium should try to represent the way Solr works, where possible. 
The components currently placed in the select query namespace as (as far as I know) only usable as select query components, so I'm not really seeing the advantage of moving them out of the select query. 

Could you explain some more what you find confusing about the current structure, and how this proposed structure improves that?

Issue #307, opened 2015-01-20
ZendHttp Adapter doesn't work with Zend Framework 2


marinpurgar:
when ZendHttp Adapter is set for the client:

$client->setAdapter('Solarium\Core\Client\Adapter\ZendHttp');

the following fatal error occurs:

Fatal error: Class 'Zend_Http_Client' not found in vendor/solarium/solarium/library/Solarium/Core/Client/Adapter/ZendHttp.php on line 153

marinpurgar:
I have changed stuff from ZendHttp.php to be ZF2 compliant and created Zend2Http.php, The PR is #308.

basdenooijer:
A new release has been done that includes a ZF2 adapter.

PR #308, opened 2015-01-20
#307 Add the Zend2Http ZF2 Client Adapter

marinpurgar:
- use \Zend\Http\Client
- use plain 'GET', 'POST' and 'HEAD' instead of \Zend_Http_Client::
- use send() instead of request()
- use isClientError() instead of isError()
- use getReasonPhrase() instead of getMessage()
- use getStatusCode() instead of getStatus()
- new headers handling, send StatusLine as the only header


basdenooijer:
Thanks for your PR, nice to have ZF2 http client integration for Solarium.
Could you please set the target merge branch to develop?

marinpurgar:
OK, I'll retarget to develop and add the PHPUnit test.

PR #321, opened 2015-03-17
Combining Stats Component With Pivots

wickedOne:
added ability to add stats to a pivot facet:
https://cwiki.apache.org/confluence/display/solr/Faceting#Faceting-CombiningStatsComponentWithPivots


.gitignore
```diff
@@ -2,4 +2,5 @@ build
 phpunit.xml
 composer.phar
 composer.lock
-vendor
\ No newline at end of file
+vendor
+nbproject/
```
basdenooijer: Not a big issue, but editor excludes should be in your global gitignore so the project only has project specific ignores.

basdenooijer:
Very nice work!

wickedOne:
thanks, very nice solr5 feature as well :-)
removed the editor ignore…

PR #323, opened 2015-04-16
Update path to autoload.php for examples/ini.php

ismaeil:
Hi,
The examples in example folder use ini.php that includes the autoload file, but the path for it is not accurate no more.

### DIFF
--- a/examples/init.php
+++ b/examples/init.php
@@ -3,7 +3,7 @@
 error_reporting(E_ALL);
 ini_set('display_errors', true);
 
-require __DIR__.'/../vendor/autoload.php';
+require __DIR__.'/../../../../vendor/autoload.php';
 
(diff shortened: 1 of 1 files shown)


basdenooijer:
This path depends on the use case. The current path is correct for a 'standalone' solarium library. Your updated path matches when Solarium is installed as a Composer dependency.

I think both should be supported, but I need to look into what is the best way to do that.

mkalkbrenner:
@ismaeil sorry for the late reply.

I think we should modify the code to rely on the composer driven autoloading.

clemens-tolboom:
> I think we should modify the code to rely on the composer driven autoloading.

- Guess we can close this PR.

mkalkbrenner:
I think, the commit of #659 addressed it

Issue #320, opened 2015-03-16
Collations format changed in Solr 5.0


eostis:
getSpellcheck()->getCollations() does not work anymore with Solr 5.0

Response from Solr 5.0:
`{"response":{"numFound":0,"start":0,"maxScore":0.0,"docs":[]},"spellcheck":{"suggestions":["cassandr",{"numFound":1,"startOffset":0,"endOffset":8,"origFreq":0,"suggestion":[{"word":"cassandra","freq":13}]}],"correctlySpelled":false,"collations":["collation",{"collationQuery":"cassandra","hits":0,"misspellingsAndCorrections":["cassandr","cassandra"]}]}}`

Response from Solr 4.xxx:
`{"response":{"numFound":0,"start":0,"maxScore":0.0,"docs":[]},"spellcheck":{"suggestions":["cassandr",{"numFound":1,"startOffset":0,"endOffset":8,"origFreq":0,"suggestion":[{"word":"cassandra","freq":13}]},"correctlySpelled",false,"collation",["collationQuery","cassandra","hits",0,"misspellingsAndCorrections",["cassandr","cassandra"]]]}}`

basdenooijer:
Thanks for reporting, we will need to add support for this new format while also keeping support for the 4.x format.

eostis:
My workaround:

	        // New 5.0 format
	        if ( isset( $data['spellcheck']['collations'] ) ) {
		        // Convert object to array
		        foreach ( $data['spellcheck']['collations'][1] as $key => $value ) {
			        $collations[] = $key;
			        $collations[] = $value;
		        }
		        $collations = $this->parseCollation($query, $collations );
	        }

	        return new SpellcheckResult\Result($suggestions, $collations, $correctlySpelled);

BBrunekreeft:
Thanks for this workaround.
Same issue with 'correctlySpelled'
"The spellcheck response format has changed, affecting xml and json clients. In particular, the "correctlySpelled" and "collations" subsections have been moved outside the "suggestions" subsection, and now are directly under "spellcheck". See SOLR-3029 for more information."
http://lucene.apache.org/solr/5_0_0/changes/Changes.html#v5.0.0.upgrading_from_solr_4.x

basdenooijer:
The fix for this issue has been included in the latest release.

PR #325, opened 2015-04-17
Spellcheck solr5

GregEisenman:
Fix for Solr5 changes to collations and correctlySpelled. Neither are currently working in Solarium and Solr5.

The code to parse Solr4 responses was untouched. This simply comes along afterwords and checks for the Solr5 locations and processes them.

#320


basdenooijer:
Thanks for your work! It will be included in the next release.

PR #327, opened 2015-04-22
Solarium bugs with Solr 5.1

bpolaszek:
Solr 5.1 now systematically returns a "Bad contentType for search handler :text/xml; charset=UTF-8" message.

It looks like a default content-type is hardcoded in Solarium library. Removing this settings solves the bug.

Not tested on former versions.

### DIFF
--- a/library/Solarium/Core/Client/Adapter/Curl.php
+++ b/library/Solarium/Core/Client/Adapter/Curl.php
@@ -155,10 +155,6 @@ class Curl extends Configurable implements AdapterInterface
             curl_setopt($handler, CURLOPT_PROXY, $proxy);
         }
 
-        if (!isset($options['headers']['Content-Type'])) {
-            $options['headers']['Content-Type'] = 'text/xml; charset=utf-8';
-        }
-
         // Try endpoint authentication first, fallback to request for backwards compatibility
         $authData = $endpoint->getAuthentication();
         if (empty($authData['username'])) {
--- a/library/Solarium/Core/Client/Adapter/Http.php
+++ b/library/Solarium/Core/Client/Adapter/Http.php
@@ -121,8 +121,6 @@ class Http extends Configurable implements AdapterInterface
                         'content',
                         $data
                     );
-
-                    $request->addHeader('Content-Type: text/xml; charset=UTF-8');
                 }
             }
         }
--- a/library/Solarium/Core/Client/Adapter/PeclHttp.php
+++ b/library/Solarium/Core/Client/Adapter/PeclHttp.php
@@ -178,9 +178,6 @@ class PeclHttp extends Configurable implements AdapterInterface
                     );
                 } else {
                     $httpRequest->setBody($request->getRawData());
-                    if (!isset($headers['Content-Type'])) {
-                        $headers['Content-Type'] = 'text/xml; charset=utf-8';
-                    }
                 }
                 break;
             case Request::METHOD_HEAD:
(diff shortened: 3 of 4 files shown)


vitaliy-w:
@bpolaszek Were you able to resolve this in a different way?

bpolaszek:
Actually, I closed this pull request as my code wasn't fully tested, and a little buggy.

After what I managed to fix this on the cUrl adapter, but I only tested with this adapter, and with Solr 5.1, not the earlier versions. I won't have the time for this... Everything's fine for me now.

I don't know if I should send a pull request with my fix since it's not fully tested...

vitaliy-w:
Thanks @bpolaszek, we've got the same issue and went with a similar short term workaround. Was curious to know how you've resolved it and it sounds like we've resolved it in similar ways.

bpolaszek:
The _\Solarium\Core\Client\Adapter\Curl::createHandle()_ method now looks like this: 
```php
    public function createHandle($request, $endpoint)
    {
        // @codeCoverageIgnoreStart
        $uri = $endpoint->getBaseUri() . $request->getUri();
        $method = $request->getMethod();
        $options = $this->createOptions($request, $endpoint);

        $handler = curl_init();
        curl_setopt($handler, CURLOPT_URL, $uri);
        curl_setopt($handler, CURLOPT_RETURNTRANSFER, true);
        if (!ini_get('open_basedir')) {
            curl_setopt($handler, CURLOPT_FOLLOWLOCATION, true);
        }
        curl_setopt($handler, CURLOPT_TIMEOUT, $options['timeout']);
        curl_setopt($handler, CURLOPT_CONNECTTIMEOUT, $options['timeout']);

        if ($proxy = $this->getOption('proxy')) {
            curl_setopt($handler, CURLOPT_PROXY, $proxy);
        }

        if (!isset($options['headers'])) {
            $options['headers'] =   [];
        }

        // Try endpoint authentication first, fallback to request for backwards compatibility
        $authData = $endpoint->getAuthentication();
        if (empty($authData['username'])) {
            $authData = $request->getAuthentication();
        }

        if (!empty($authData['username']) && !empty($authData['password'])) {
            curl_setopt($handler, CURLOPT_USERPWD, $authData['username']. ':' . $authData['password']);
            curl_setopt($handler, CURLOPT_HTTPAUTH, CURLAUTH_BASIC);
        }

        if ($method == Request::METHOD_POST) {

            if (!isset($options['headers']['Content-Type'])) {
                $options['headers']['Content-Type'] = 'text/xml';
            }

            curl_setopt($handler, CURLOPT_POST, true);

            if ($request->getFileUpload()) {
                if (version_compare(PHP_VERSION, '5.5.0') >= 0) {
                    $curlFile = curl_file_create($request->getFileUpload());
                    curl_setopt($handler, CURLOPT_POSTFIELDS, array('content', $curlFile));
                } else {
                    curl_setopt($handler, CURLOPT_POSTFIELDS, array('content' => '@'.$request->getFileUpload()));
                }
            } else {
                curl_setopt($handler, CURLOPT_POSTFIELDS, $request->getRawData());
            }

        } elseif ($method == Request::METHOD_GET) {
            curl_setopt($handler, CURLOPT_HTTPGET, true);
        } elseif ($method == Request::METHOD_HEAD) {
            curl_setopt($handler, CURLOPT_CUSTOMREQUEST, 'HEAD');
        } else {
            throw new InvalidArgumentException("unsupported method: $method");
        }

        if (count($options['headers'])) {
            $headers = array();
            foreach ($options['headers'] as $key => $value) {
                $headers[] = $key . ": " . $value;
            }
            curl_setopt($handler, CURLOPT_HTTPHEADER, $headers);
        }

        return $handler;
        // @codeCoverageIgnoreEnd
    }
```

As you can see, I just init a _$options['headers']_ array if no header was provided, to avoid _E_NOTICE_ errors. 
Then, only in the POST method, if the content-type wasn't set to _text/xml_, let's do it now. I had to remove _charset=utf8_ too.
The piece of code that populates the headers have been moved to the bottom of the method, after the _if ($method == Request::METHOD_POST)_ condition.

PR #328, opened 2015-04-23
Update Curl.php

dagnew:
Allow GET requests such as /select to validate in SOLR 5.1

### DIFF
--- a/library/Solarium/Core/Client/Adapter/Curl.php
+++ b/library/Solarium/Core/Client/Adapter/Curl.php
@@ -156,7 +156,7 @@ class Curl extends Configurable implements AdapterInterface
         }
 
         if (!isset($options['headers']['Content-Type'])) {
-            $options['headers']['Content-Type'] = 'text/xml; charset=utf-8';
+            $options['headers']['Content-Type'] = 'application/x-www-form-urlencoded; charset=utf-8';
         }
 
         // Try endpoint authentication first, fallback to request for backwards compatibility


rmzamora:
This patch will only work if your doing GET but will throw an error for POST.

dagnew:
In the end I went with:
      if (!isset($options['headers']['Content-Type'])) {
            if($method == 'GET'){
                $options['headers']['Content-Type'] = 'application/x-www-form-urlencoded; charset=utf-8';
            }else{
                $options['headers']['Content-Type'] = 'application/xml; charset=utf-8';
            }    
        }

thomascorthals:
dagnew's code in the previous comment works for me as well. But to keep with the coding style, I suggest coding the check as ($method == Request::METHOD_GET) instead.

lammensj:
I can confirm that the code of @dagnew works for me. It would be great if this gets merged.

mfpierre:
If you're using NelmioSolariumBundle you can override the client adapter_class with a custom Curl class.

Here is an updated version of @dagnew's code
```php
if (!isset($options['headers']['Content-Type'])) {
    if($method == Request::METHOD_GET){
        $options['headers']['Content-Type'] = 'application/x-www-form-urlencoded; charset=utf-8';
    } else {
        $options['headers']['Content-Type'] = 'application/xml; charset=utf-8';
    }
}
```

basdenooijer:
This PR still has the code that only works for GET requests, not the code mentioned in the comments that works for both GET and POST. 
If this gets updated and we can also confirm this works for solr < 5, I will merge it and create a new release.

mfpierre:
I just made a new PR with the updated code to make this merged as quick as possible.

I can confirm on my side that it's working as well (solr 5.1)

basdenooijer:
Closing this PR in favor of #334

PR #330, opened 2015-05-13
Added support for facet.contains and facet.contains.ignoreCase

thomascorthals:
Sorl 5.1.0 added facet.contains and facet.contains.ignoreCase. I've added contains() and containsIgnoreCase() methods to facetsets and fields.

I had to remove the default Content-Type in the Curl adapter because the newer Solr doesn't accept text/xml anymore.


thomascorthals:
My first tests where against an index that wasn't built from my application. It turns out my changes to the Curl adapter only work when searching, not when updating. dagnew's comment in pull request #328 has code that does work in both cases.

PR #334, opened 2015-06-10
Fix curl headers for solr > 5

mfpierre:
Updated version of #328 thanks to @dagnew for making the original PR.

PR #336, opened 2015-06-11
Fix curl headers for solr > 5

mfpierre:
#334 but on the right branch :disappointed:


basdenooijer:
Thanks! I will try to create a new release soon.

PR #349, opened 2015-07-06
nested document support

aminshoman:
Solr supported nested documents starting from version 4.7
http://yonik.com/solr-nested-objects/

My change simply embed sub-arrays into a multi-level document xml structure:

``` PHP
        $update = $client->createUpdate();
        $doc1 = $update->createDocument();
        $doc1->id = uniqueid();
        $doc1->title = 'I am parent';
        $doc1->child_items = array( 
            array(
                 'id' =>uniqueid(),
                 'title' =>'I am child 1',
                  ),
            array(
                 'id' =>uniqueid(),
                 'title' =>'I am child 2',
                  ),
        );
```

and the xml output will be:

``` XML
<add>
   <doc>
       <field name="id">123</field>
       <field name="title">I am parent</field>
         <doc>
           <field name="id">456</field>
           <field name="title">I am child 1</field>
      </doc>
     <doc>
         <field name="id">789</field>
         <field name="title">I am child 2</field>
     </doc>
   </doc>
</add>
```


basdenooijer:
I think this would be a great addition, but currently the PR is cluttered with other changes in the curl adapter that cannot be merged (for instance YII debug statements)

Maybe the best option is to create a new, clean, PR with just the changes to the RequestBuilder. Also some unittests should be added to cover this new functionality. Are you willing to do this? Otherwise I could also create a new PR based on your ideas.

daoth:
After working with solariumphp library in some days
I feel very pity because library dosen't support nested document
Omg :(

thePanz:
Looks like this feature has been implemented in https://github.com/solariumphp/solarium/commit/b6c66a8696ffabaeb63d4e25e3aa28fd011538dc

An available in Solarium v3.8.1

Issue #351, opened 2015-07-09
Setting CURL-adapter to explicitly use no proxy


Ancillius:
I have to avoid using the system wide proxy for my solarium script, so i tried to do the following:

    $client = new Solarium\Client();
    $client->setOptions([
      'adapteroptions' => [
        'proxy' => '',
      ],
    ]);

Setting the solarium adapter proxy options  to "" (empty string) does not work, since Solarium\Core\Client\Adapter : 154 checks:

        if ($proxy = $this->getOption('proxy')) {
            curl_setopt($handler, CURLOPT_PROXY, $proxy);
        }

Thus an empty string would be casted to "false" and would not be set as proxy.

What about changing it to 

        if (($proxy = $this->getOption('proxy')) !== null) {
            curl_setopt($handler, CURLOPT_PROXY, $proxy);
        }

This helps in my case, but does it break anything?

basdenooijer:
A fix for this has been merged in the develop branch

PR #358, opened 2015-08-14
Make it possible to bypass proxy.

keesschepers:
We have the exact same issue as #351 and therefor I implemted his change which seems to work quite well for us.


basdenooijer:
Thanks!

PR #370, opened 2015-11-30
Added support for facet.contains and facet.contains.ignoreCase

basdenooijer:
Merge of https://github.com/solariumphp/solarium/pull/330/files

Also added test coverage


basdenooijer:
Wrong target

PR #377, opened 2015-11-30
add php 7.0 to travis manifest

shieldo:
PHP 7.0 final is expected to be released in the coming days - so there should be non-failing tests against it.


basdenooijer:
I agree, thanks for the PR!

PR #390, opened 2016-01-14
added support for replicas in distributed search

wickedOne:
for load balancing purposes as described @ https://cwiki.apache.org/confluence/display/solr/Distributed+Requests

added facet interval test as a bonus :sunglasses:


basdenooijer:
Nice work, even including updated docs!

PR #398, opened 2016-02-16
Stats percentiles

eborned:
Support for fetching additional stats values, like percentiles.
Added existence checks before fetching values from stats array.


basdenooijer:
Thanks!

PR #404, opened 2016-03-03
docs(queries): fixes php docblocks without preceding line break

calumbrodie:
- These docsblocks do not render correctly on read the docs unless the code block is on a new line

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -37,13 +37,19 @@ See [<https://packagist.org>](https://packagist.org) for other packages.
 
 - Make sure to use the composer autoloader, and Solarium should be available.
 
-**Only if you don't use composer:** you need to use a PSR-0 autoloader, or the supplied autoloader: ```php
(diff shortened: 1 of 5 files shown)


docs/getting-started.md
```diff
@@ -37,13 +37,19 @@ See [<https://packagist.org>](https://packagist.org) for other packages.
...
-**Only if you don't use composer:** you need to use a PSR-0 autoloader, or the supplied autoloader: ```php
+**Only if you don't use composer:** you need to use a PSR-0 autoloader, or the supplied autoloader:
+
+```php
 
 Also you need to make sure the the Symfony Event Dispatcher component is available.
 
 ### Checking the availability
 
+<<<<<<< e523aa92f2d3d0b9aac6764fd3b35e70bf1369ea
```
basdenooijer: Seems like a conflict has not been resolved? The same for several lines below.

basdenooijer:
Looks like a good change, but there is one file which contains some conflicts.

Issue #411, opened 2016-03-30
Curl adapter does not return HTTP response headers


cristiroma:
The Curl adapter does not parse & return the HTTP response headers. Sometimes the headers returned by Solr could be useful to inspect.

PR #412, opened 2016-03-30
#411 Add HTTP response headers to the Curl adapter

cristiroma:
Regarding #411 changes overview
- Changed HTTP/1.1 to HTTP 1.0, and forced curl to use HTTP 1.0 protocol, I'm not sure why 1.1 is required, but the response header was not correctly populated with 1.0
- I have configured curl to return the response headers and I've made a function to parse them then attach to the response
- Wrote a test using a configurable Solar to test the response.


library/Solarium/Core/Client/Adapter/Curl.php
```diff
@@ -256,4 +261,36 @@ protected function createOptions($request, $endpoint)
...
+    {
+        $headers = array();
+        // Split the string on every "double" new line.
+        $arrRequests = explode("\r\n\r\n", $content);
+        // Loop of response headers. The "count() -1" is to
+        //avoid an empty row for the extra line break before the body of the response.
+        for ($index = 0; $index < count($arrRequests) -1; $index++) {
+            foreach (explode("\r\n", $arrRequests[$index]) as $i => $line) {
+                if ($i !== 0) { // First is HTTP 1.0 200 OK.
+                    list ($key, $value) = explode(': ', $line);
```
cristiroma: It's not clear how the returned array should look like. Looking at acquia_connector, it appears that is assuming to be just strings: http://cgit.drupalcode.org/acquia_connector/tree/acquia_search/src/EventSubscriber/SearchSubscriber.php?h=8.x-1.x#n97

Do you have any idea?

```php
$headers = array('Expires:Fri, 01 Apr 2016 20:41:50 GMT', 'Keep-Alive:timeout=2, max=100')
``` 

or 

```php
$headers = array('Expires' => 'Fri, 01 Apr 2016 20:41:50 GMT', 'Keep-Alive'  => 'timeout=2, max=100');
```

cristiroma:
I have investigated the issue and apparently the ``Http`` Adapter is returning headers as array of strings: https://github.com/solariumphp/solarium/blob/c5fc5255a926abdb2a682bf0fb2dea4c3ebf413a/library/Solarium/Core/Client/Adapter/Http.php#L175 - so I modified the ``Curl`` adapter to do the same

PR #413, opened 2016-03-30
301 malformed multipart form upload

cristiroma:
Added fix for error 

```
SEVERE: null:org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found
```

Also a test to showcase the error


janusman:
So, I have what I think is a working PR at https://github.com/solariumphp/solarium/pull/484 ... if that is sane, any way we could add my changes to this PR?

cristiroma:
To anyone coming here we are using this patch half a year and it works just fine. _I enabled edit from maintainers_, hopefully someone can help me writing a test. I don't exactly understand how to write make Travis work.

PR #419, opened 2016-05-11
Make sure floats use "." decimal separator

mpdude:
The current locale applies when casting floats to strings. For example, with the `de_DE` locale, a boost value of 12.34 will be case to `"12,34"`. This causes a downstream exception in Solr because it does not understand this format.

`number_format` is a bit hacky but is independent of the current locale.

### DIFF
--- a/library/Solarium/QueryType/Update/RequestBuilder.php
+++ b/library/Solarium/QueryType/Update/RequestBuilder.php
@@ -125,7 +125,7 @@ class RequestBuilder extends BaseRequestBuilder
 
         foreach ($command->getDocuments() as $doc) {
             $xml .= '<doc';
-            $xml .= $this->attrib('boost', $doc->getBoost());
+            $xml .= $this->attrib('boost', number_format($doc->getBoost(), 5, '.', ''));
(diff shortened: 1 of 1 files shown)


mpdude:
Rebased against master, re-running tests

thomascorthals:
You can lose precision with `number_format` if your boost has more decimals than you are formatting it with.

Using `setlocale(LC_NUMERIC, 'C')`is IMHO a better solution for this issue. Just make sure you restore it to whatever the user had set afterwards. I've created a patch against the current master with a working example: https://gist.github.com/thomascorthals/588ff04ff44b3f568de3b4809f787b54

mkalkbrenner:
The PR is outdated for two reasons if you look at the current code:
```php
    /**
     * Get the document boost value.
     *
     * @return float|null
     *
     * @deprecated No longer supported since Solr 7
     */
    public function getBoost(): ?float
    {
        return $this->boost;
    }
```

1. The return type of the function is declared to be a float.  In former times it might have returned a string.
2. The function is deprected because Solr itself removed the feature of per document boosts.

PR #421, opened 2016-05-18
Added support for deep paging with a cursor

thomascorthals:
I've added the possibility to use Solr's cursor functionality for efficient deep paging. PrefetchIterator makes use of this implementation if a cursorMark is set for the query.


basdenooijer:
Thanks a lot for creating this PR, a nice feature to add to Solarium! 

The implementations looks good, but before merging this I think we should look into creating some test coverage and basic documentation for this feature.

majun:
Hi, how it looks with this feature?

PR #424, opened 2016-05-24
Remove Client dependency from Result

ribeiroplinio:
I'm using the library and caching the Solr Result for some cases and I had some errors regarding serializing the Result object. The serialize problem itself is related to the Event Dispatcher dependency on Client https://github.com/solariumphp/solarium/blob/master/library/Solarium/Core/Client/Client.php#L551.

Anyway, the Client dependency on Result doesn't make much sense for me, and it's not used. That's the reason for the PR.

### DIFF
--- a/library/Solarium/Core/Client/Client.php
+++ b/library/Solarium/Core/Client/Client.php
@@ -739,7 +739,7 @@ class Client extends Configurable implements ClientInterface
         }
 
         $resultClass = $query->getResultClass();
-        $result = new $resultClass($this, $query, $response);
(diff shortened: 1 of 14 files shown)


marinpurgar:
:+1:

basdenooijer:
I agree the client shouldn't be included in the result. In the library itself I cannot find any usages of the client instance in the result objects, however removing it is a BC break so should be done in a major release.
For now I'll keep this PR open, I hope to start on a major release soon.

PR #429, opened 2016-06-03
Accept DateTimeImmutable objects in helper

bpolaszek:
This pull request aims at accepting any `DateTimeInterface` / `DateTimeImmutable` objects in the query helper that transform date objects to Solr UTC ISO8601 date format.

Since these native PHP interface / object have been introduced in PHP 5.5, `composer.json` needs to be modified.

The whole library shouldn't force PHP version upgrade just for that small feature, so there's no emergency at all.  It's just for getting it in the pipe for the next major version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.2",
+        "php": ">=5.6",
         "symfony/event-dispatcher": "~2.3|~3.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


basdenooijer:
Thanks, nice improvement. I'm not sure yet what the version requirement is going to be for the next major version. I'd like to go for only versions that are officially supported (which will be 5.6+ in about a month) but the reality is a lot of people are far behind on upgrades.

bpolaszek:
Thank you @mkalkbrenner,

I actually did not remember I did that 🙃

PR #431, opened 2016-06-06
Solr allows use key and ex

vlastv:
Tested on solr 5 and 6 versions


vlastv:
Fix tests?

vlastv:
Proof https://issues.apache.org/jira/browse/SOLR-7019

vlastv:
Link with #448

basdenooijer:
Thanks!

Issue #420, opened 2016-05-15
Solr Criteria Query Builder


dewos:
Hello there. As has already been mentioned in #123 (@basdenooijer), I would like to help the project with a fluent query builder. I made an implementation of Spring Solr Criteria for PHP, and soon I'll tag a stable release. You can find it here: https://github.com/minimalcode-org/search

You could "suggest" it (composer or documentation) or use it directly as the official builder, I don't care for the credit. My advice actually is to do something like aliasing, for keeping the namespace in Solarium context:

```php
<?php
// solarium/library/Solarium/Core/Query/Criteria.php
namespace namespace Solarium\Core\Query;
use Minimalcode\Search\Criteria as BaseCriteria;

class Criteria extends BaseCriteria {}

```
I hope it is useful, and I'm open to comments and suggestions, thanks :+1:

basdenooijer:
Nice work, it looks interesting 👍 

I think your library can be used with Solarium perfectly without a special integration. We can suggest it in the docs, and maybe include an example. Anyone who wants to use it can easily do so using composer, and we could also consider to add it to the composer file as a suggested package.

I don't think we need to create a namespace alias, people are getting used to the trend of small libraries for a specific purpose, instead of big monolithic libraries.

dewos:
I wanted to gift it to the solarium, but a suggestion in the doc with some examples is also very good idea @basdenooijer. I'm committed to keeping the repo online and alive anyway. 

And yes, luckly the PHP community has evolved very much in recent years, maybe I could contact the other "Solr-Php libraries" (FloriansemmSolrBundle etc..) for extending the invitation on them too.

See you next time, and congratulations for the project. :+1:

basdenooijer:
Great, next time I'm editing the docs I'll try to remember to add this. Or, if you want to, you can also send a PR with updated docs :)

bpolaszek:
Hello Dewos,

Your work looks great !
Do you use the Doctrine library ? 

Their code looks very similar: http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/working-with-associations.html#filtering-collections

The big advantage of their "criteria" implementation is that it's not coupled to SQL filtering, even not to the Doctrine ORM itself.

You can grab the standalone library here: https://github.com/doctrine/collections

Doctrine currently has 2 implementations:
* Transforming criteria expressions into SQL syntax
* Transforming criteria expressions into PHP (for filtering an array, for instance).

[...]

PR #434, opened 2016-06-27
[Proposal] solr query-builder

floriansemm:
Related issue #420 

QueryBuilder allows to define queries programmatically.

```php
$builder
    ->select('title', 'content')
    ->where('id')
        ->equals('id_prefix_*')
    ->andWhere('title_s')
        ->equals('a title')
    ->orWhere('tags_ss')
        ->inSet(array('solr', 'symfony'));
```

A condition starts always with `where` / `andWhere` / `orWhere`, followed by `equals($value)`/ `inSet($values)`/ `inRange($start, $end)`/ `inDistance($lat, $long, $distance)`.

I have also implemented a test-class to show the usage.

The implementation and featureset is very basic and should only start a discussion.

### DIFF
--- a/library/Solarium/Core/Query/QueryBuilder.php
+++ b/library/Solarium/Core/Query/QueryBuilder.php
@@ -1,0 +1,206 @@
+<?php
+
+namespace Solarium\Core\Query;
+
+use Solarium\Client;
+use Solarium\QueryType\Select\Query\Query;
+
+class QueryBuilder
+{
+    /**
+     * @var array
+     */
+    private $queryParts = array();
+
+    /**
+     * @var Client
+     */
+    private $client;
+
+    /**
+     * @var array
+     */
+    private $selectFields;
+
+    /**
+     * @var AbstractQuery
+     */
+    private $query;
+
+    /**
+     * @var bool
+     */
+    private $buildingWhere;
+
+    /**
+     * @param Client $client
+     */
+    public function __construct(Client $client)
+    {
+        $this->client = $client;
+    }
+
+    /**
+     * @param string $field
+     *
+     * @return QueryBuilder
+     */
+    public function where($field)
+    {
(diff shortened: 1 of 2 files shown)


dewos:
Interesting, but it is a big step backwards (features-wise) compared with #420, which has not even been merged neither. Probably it would be better to focalize the efforts into one single lib.

basdenooijer:
I think it's clear there is demand for a query builder, and I can also see the benefits myself. 

As already discussed in #420 I think something like this could be a separate library, as it's very useful even without Solarium. 
Since it is related to Solarium, we could create a project under the 'solariumphp' github organization.

dewos:
@basdenooijer merge, solariumphp/builder, community lib, composer suggestion, .... the important thing is to do.... something :)

basdenooijer:
I don't want to force any of the options. But the way I see it:
- The community lib option is always open to anyone.
- I'm also willing to merge a PR that adds such a lib to the composer suggestions
- If someone wants to take ownership of solariumphp/builder (or whatever name) I'm open for that.

Currently I don't have the time to take the lead in such a project, but if someone wants to do that (or even better, several people?) please let me know.

dewos:
Hi @basdenooijer . Now that much we have all agreed upon, that the PHP community would benefit from a query builder. 

The simplest and fastest solution is to just pursue your idea of #420 to sponsorize the stand-alone library (@floriansemm could do the same on the bundle). Just that would help a lot, and it can happen tomorrow. 

A solariumphp/builder repository is also an equivalent solution. The important thing is to direct the community towards an agreed solution. Speaking of which: probably there is not so much else to do to Criteria, features wise.

Like I said it is the result of many years of java-community brainstorming, and in my opinion is a very good solution for 90% of the use-case. It would be a good idea to keep the API as it is right now, at least initially, to facilitate developers that switch between the two languages.

PR #438, opened 2016-06-30
Do not parse collation array for Solr 5+; values are already in name/…

haydenyoung:
…value pairs. Change unit test to match the collation data structure returned by Solr.

PR #439, opened 2016-07-01
Do not parse collation array for Solr 5+; values are already in name/value pairs. Change unit test to match the collation data structure returned by Solr.

haydenyoung:
@basdenooijer No problem. PR against develop branch. Replaces #438.


basdenooijer:
Thanks!

Issue #440, opened 2016-07-27
Empty grouping results when using group.func


pavelsmolka:
When I use a grouping parameter `group.func`, calling `$result->getGrouping()->getGroups()` returns an empty array.

The underlying reason seems to be that `Solarium\QueryType\Select\ResponseParser\Component\Grouping` checks the Grouping fields in the response data. However, if the function is used, no grouping fields are set, and therefore the data is discarded.

It seems that checking for `$grouping->getFunction()` in the `parse` method of the mentioned class would fix this. If you can confirm my thinking is right, I can provide a pull request.

basdenooijer:
I think that would be the correct solution. Currently only grouping fields and queries are checked, a function check should be added. I think large parts of the code for field group parsing could be reused (by abstracting them first).

basdenooijer:
This has been resolved in #441

PR #441, opened 2016-07-31
Add support for grouping result using a function

pavelsmolka:
Add functionality that checks whether the result data in
ResponseParser\Component\Grouping contain groups created by
solr parameter `group.func` (an arbitrary Solr-supported
function).

When such field is found, apply the same parsing logic
which is usually used for field-grouping, in order to ensure
the function-grouping is recognized and added to the FieldGroup
result object.

This should add a new functionality requested in ticket #440

### DIFF
--- a/library/Solarium/QueryType/Select/ResponseParser/Component/Grouping.php
+++ b/library/Solarium/QueryType/Select/ResponseParser/Component/Grouping.php
@@ -67,7 +67,9 @@ class Grouping implements ComponentParserInterface
             // parse field groups
             $valueResultClass = $grouping->getOption('resultvaluegroupclass');
             $documentClass = $query->getOption('documentclass');
-            foreach ($grouping->getFields() as $field) {
+
+            // check grouping fields as well as the grouping function (either can be used in the query)
+            foreach (array_merge($grouping->getFields(), array($grouping->getFunction())) as $field) {
                 if (isset($data['grouped'][$field])) {
                     $result = $data['grouped'][$field];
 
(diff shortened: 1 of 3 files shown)


library/Solarium/QueryType/Select/ResponseParser/Component/Grouping.php
```diff
@@ -67,7 +67,9 @@ public function parse($query, $grouping, $data)
             // parse field groups
             $valueResultClass = $grouping->getOption('resultvaluegroupclass');
             $documentClass = $query->getOption('documentclass');
-            foreach ($grouping->getFields() as $field) {
+
+            // check grouping fields as well as the grouping function (either can be used in the query)
+            foreach (array_merge($grouping->getFields(), [$grouping->getFunction()]) as $field) {
```
basdenooijer: Currently PHP 5.3 is still supported, so short array syntax must be avoided. The next major Solarium release will update the requirements.

basdenooijer: I just see the current CI config allows 5.3 failures, that's why the tests pass. But the composer file still requires 5.3.2 and I cannot change that in a minor release.

basdenooijer:
Thanks for you fix, apart from one small issue regarding short array syntax the changes look good. Could you fix the array, and retarget the PR to the develop branch? Then I will merge it.

pavelsmolka:
I have fixed the array syntax and rebased the branch - it should be ready to get merged now.

basdenooijer:
Thanks!

PR #443, opened 2016-08-12
Fix fenced block start.

krtek4:
Newlines are missing, resulting in the code not correctly highlighted in the documentation.


basdenooijer:
Thanks for fixing!

PR #449, opened 2016-09-15
Fix interface typehint in docblock

bramstroker:
Update document interface should be considered instead of select documents for the BufferedAdd plugin


basdenooijer:
Thanks Bram!

PR #453, opened 2016-10-17
add spatial component

kaczenski:
We added a spatial component in order to put solr spatial params ("pt", "d" and "sfield") onto the request.

```
// create a client instance
$client = new Solarium\Client($config);

// get a select query instance
$query = $client->createSelect();

// get the dismax component and set a boost query
$spatial = $query->getSpatial();
$spatial->setPoint('52,13');
$spatial->setField('geo');
$spatial->setDistance(50);

// this executes the query and returns the result
$resultset = $client->select($query);
```

Herewith you can use {!geofilt} as fq.
Also geodist() asc works out of the box for sorting without adding a field and calculating the score. 

Further reading:
https://cwiki.apache.org/confluence/display/solr/Spatial+Search

### DIFF
--- a/library/Solarium/QueryType/Select/Query/Component/Spatial.php
+++ b/library/Solarium/QueryType/Select/Query/Component/Spatial.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Solarium\QueryType\Select\Query\Component;
+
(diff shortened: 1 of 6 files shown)


basdenooijer:
Thanks!

PR #464, opened 2016-12-15
 Add Guzzle3 Client Adapter

chadicus:
This pull request adds a Guzzle3 adapter. In order to maintain this library's PHP 5.3 compatibility I am choosing to implement v3 of guzzle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
         "phpunit/phpunit": "~3.7",
         "squizlabs/php_codesniffer": "~1.4",
         "zendframework/zendframework1": "~1.12",
-        "satooshi/php-coveralls": "~1.0"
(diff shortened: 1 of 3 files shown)


basdenooijer:
Nice addition, thanks! 

For the next major version of Solarium I'm even considering to replace all adapters by a single library, with Guzzle as the primary candidate, but not sure on this yet. With that in mind it's good to start with an adapter implementation for Guzzle.

chadicus:
@basdenooijer I think componentizing the adapters would be a great idea for the next major version.  

What are the plans for the release of this PR?  Will it be released with version 3.8 or with a patch to v3.7?

basdenooijer:
This will be included in the next minor version

PR #466, opened 2016-12-16
Add Guzzle 6 Adapter implementation

chadicus:
This pull request adds a Guzzle 6 Adapter implementation.

To avoid backwards breaking compatiblity the composer.json file was changed to allow for guzzle 3 or 6. The unit tests will also be marked as skipped depending on which version is installed.


chadicus:
@basdenooijer this PR is no longer backwards breaking

basdenooijer:
Nice work!

PR #467, opened 2016-12-17
usleep after curl_multi_select -1

bjoernhaeuser:
According to these php bugs:

* https://bugs.php.net/bug.php?id=61141
* https://bugs.php.net/bug.php?id=63842

thats what we should do.

I came across this issue when trying to make the ParallelExecution
working on my CentOS 7.3 box. It spinlooped inside the changed loop.

Also guzzle is doing the same thing:

https://github.com/guzzle/guzzle/blob/master/src/Handler/CurlMultiHandler.php#L103

I am not sure if the timeout needs to be configurable and if the choosen
value (100) is good.

Thank you very much for looking into the change. I do no think we can add a test for this. :/


basdenooijer:
I was not aware of this issue, but it seems like the correct solution. Thanks for your PR and the background info!

PR #469, opened 2016-12-21
Correct language about RealtimeGet query use

brazenest:
This submission is a correction to an unclear sentence in the docs about the RealtimeGet component.


basdenooijer:
Thanks!

PR #470, opened 2016-12-26
Fix document syntax

suzuki:
## Overview

- The layout of document is broken:
    - https://github.com/solariumphp/solarium/blob/0acdb2838ae551fe87e4e334e449a468973839f2/docs/queries/update-query/building-an-update-query/building-an-update-query.md


basdenooijer:
Thanks!

PR #472, opened 2016-12-27
Updated parameter names for suggestor queryType

rjcrystal:
Suggestor component uses suggest request parmeters in new solr versions.
using spellcheck parameter creates problem with suggest plugin.


mkalkbrenner:
Duplicate of #501

PR #483, opened 2017-02-15
Provides fluent interface

xcezx:


thePanz:
Before considering merging this PR, please have a read here: http://ocramius.github.io/blog/fluent-interfaces-are-evil/

basdenooijer:
Thanks for your PR, I think a fluent interface makes sense in this case

PR #485, opened 2017-02-16
Performance updates for formatting values

tonynelson19:
Adjusts the conditional logic when building the field XML to only call htmlspecialchars on unknown values.


basdenooijer:
Thanks!

Issue #492, opened 2017-04-05
Unnecessary (?) use of ini_get in Curl Adapter


ejn:
In many hosting environments, the function ini_get is disabled for security reasons. In this case the Curl Adapter creates warnings due to the attempt to check the value of open_basedir.

This could be easily fixed using a check on function_exists('ini_get'), which returns FALSE in the case that this function is disabled.

mkalkbrenner:
Since #493 is committed, we can close this one

PR #493, opened 2017-04-05
Check ini_get is not disabled before using it

ejn:
As described in #492, ini_get may often be disabled in production environments, causing warnings to be generated (and NULL) to be returned. This change checks that ini_get is usable (otherwise function_exists will return FALSE) before using it, preventing the warnings while introducing no functional change.


basdenooijer:
Thanks!

PR #496, opened 2017-04-26
fix(client): set the connection_timeout for Guzzle v3

gsdevme:
From what I see on the Guzzle v3 documentation its `connection_timeout` 
https://guzzle3.readthedocs.io/http-client/client.html?highlight=timeout#timeout-connect-timeout


basdenooijer:
thanks!

PR #497, opened 2017-04-26
fix(client): set the connection timeout as the timeout (like Guzzlev3)

gsdevme:
The Guzzle3 client sets the connection_timeout.. therefore seems fitting the Guzzle client also follows.. 

Although its less than ideal to use the same method for both connection timeout & timeout as the numbers would be different ideally.

### DIFF
--- a/library/Solarium/Core/Client/Adapter/Guzzle.php
+++ b/library/Solarium/Core/Client/Adapter/Guzzle.php
@@ -74,6 +74,7 @@ class Guzzle extends Configurable implements AdapterInterface
             RequestOptions::HEADERS => $this->getRequestHeaders($request),
             RequestOptions::BODY => $this->getRequestBody($request),
             RequestOptions::TIMEOUT => $endpoint->getTimeout(),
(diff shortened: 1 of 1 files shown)


library/Solarium/Core/Client/Adapter/Guzzle.php
```diff
@@ -74,6 +74,7 @@ public function execute($request, $endpoint)
             RequestOptions::HEADERS => $this->getRequestHeaders($request),
             RequestOptions::BODY => $this->getRequestBody($request),
             RequestOptions::TIMEOUT => $endpoint->getTimeout(),
+            RequestOptions::CONNECT_TIMEOUT => $endpoint->getTimeout()
```
shieldo: Please add a trailing comma.

gsdevme: Corrected. although as a whole this project doesn't seem valid PSR2 or Symfony Standard so I wasn't able to run phpcs

PR #501, opened 2017-05-08
Make suggester query use suggest prefix

tomgud:
This PR changes the behavior of the Suggester query that makes requests against `/suggest` to use the suggest prefixes and parameters described in [Suggester Apache Documentation](https://cwiki.apache.org/confluence/display/solr/Suggester).

The suggester query uses the spellcheck components although they should be used with the spell check endpoint. Right now my queries for anything other than the default suggester dicitonary get "redirected" to the default, because solr doesn't listen to `spellcheck.dictionary`. 

To use spell check requests the library should add a new query to meet that demand.

This may have worked with earlier versions of Solr, or I am completely missing the point on how to use a different dictionary with a suggester query in the solarium client.

### DIFF
--- a/library/Solarium/QueryType/Suggester/RequestBuilder.php
+++ b/library/Solarium/QueryType/Suggester/RequestBuilder.php
@@ -59,12 +59,10 @@ class RequestBuilder extends BaseRequestBuilder
     public function build(QueryInterface $query)
     {
         $request = parent::build($query);
-        $request->addParam('spellcheck', 'true');
-        $request->addParam('q', $query->getQuery());
-        $request->addParam('spellcheck.dictionary', $query->getDictionary());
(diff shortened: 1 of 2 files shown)


tomgud:
This is similar to https://github.com/solariumphp/solarium/pull/472 but tests should pass on this PR.

tomgud:
@basdenooijer Any chance of this getting merged anytime soon?

basdenooijer:
First of all, sorry about the delay.

I think this PR is the correct implementation, however some people might depend on the current implementation. I haven't tested, but I assume the new implementation is not backwards compatible?

basdenooijer:
Related: #472

tomgud:
@basdenooijer No problem for the delay 😄 - I think you are quite right, this is not backwards compatible so I guess it might need a version bump.

PR #503, opened 2017-06-10
Fix: Use PSR-4 autoloading

localheinz:
This PR

* [x] switches to PSR-4 autoloading
* [x] flattens the directory structure

💁‍♂️ For reference, see https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md#autoloading-standard:

>**Deprecated** - As of 2014-10-21 PSR-0 has been marked as deprecated. [PSR-4](http://www.php-fig.org/psr/psr-4/) is now recommended as an alternative.


localheinz:
Let me know when I should rebase this one, @basdenooijer!

localheinz:
Thank you, @mkalkbrenner!

PR #505, opened 2017-06-10
Fix: Cache dependencies installed with composer between builds

localheinz:
This PR

* [x] caches dependencies installed with `composer` between builds on Travis

💁‍♂️ For reference, see https://docs.travis-ci.com/user/caching/#Arbitrary-directories.


localheinz:
Thank you, @basdenooijer!

PR #506, opened 2017-06-10
Fix: Remove useless override

localheinz:
This PR

* [x] removes a useless method override


localheinz:
Thank you, @basdenooijer!

PR #509, opened 2017-06-10
Fix: Move assignment of default values before condition

localheinz:
This PR

* [x] moves the assignment of default values from an `else` branch before the condition


localheinz:
Thank you, @basdenooijer!

PR #510, opened 2017-06-10
Fix: Inline variable

localheinz:
This PR

* [x] inlines a variable


localheinz:
Thank you, @basdenooijer!

PR #512, opened 2017-06-10
Enhancement: Configure phpunit to use color

localheinz:
This PR

* [x] configures `phpunit` to use colors with its output


phpunit.xml.dist
```diff
@@ -1,5 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit backupGlobals="false" backupStaticAttributes="false" syntaxCheck="false" bootstrap="tests/bootstrap.php">
+<phpunit
+    backupGlobals="false"
+    backupStaticAttributes="false"
+    syntaxCheck="false"
+    bootstrap="tests/bootstrap.php"
+    colors="true"
+>
```
localheinz: Wrapping here because the line is already quite long.

localheinz:
Thank you, @basdenooijer!

PR #513, opened 2017-06-10
Fix: Case mismatch

localheinz:
This PR

* [x] fixes case mismatches in class and method references


localheinz:
Thank you, @basdenooijer!

PR #519, opened 2017-08-10
Fixes bugs from PR 484; fix Http adapter for extraction requests

janusman:
So, PR 484 was merged into the Develop branch, but that PR then breaks extraction requests if using the Curl adapter.

This PR focuses on only fixing the Http adapter.


janusman:
Tests actually passed--they failed only on HHVM, because travis says "HHVM is no longer supported on Ubuntu Precise".

mkalkbrenner:
The PR has conflicts against 4.x. Can you re-roll it?

mkalkbrenner:
https://travis-ci.org/solariumphp/solarium

PR #531, opened 2017-11-28
Fix: bufferAdd only allowed commitWithin/overwrite values to be set w…

mulligaj:
…hen flush

is manually called and would not set these values when the flush method automatically triggered due to reaching the buffersize.
Getter/setter methods were provided similar to bufferSize to allow these values to be preset so that automatic calls to flush
would still provide these values.


mkalkbrenner:
I re-rolled this PR in #636

PR #543, opened 2018-01-12
Allow symfony/event-dispatcher 4.x

thePanz:


mkalkbrenner:
I think, we already support solarium 4.x in the solarium 4.x branch. Can you review it?

thePanz:
@mkalkbrenner not sure, about 4.x branch of Solarium.
Having symfon/event-dispatcher-4.x supported in Solarium version 3 can help in projects using solarium and willing to completely move to Symfony4 components.
I am already doing it, by having `"symfony/event-dispatcher": "4.0.3 as 3.4.3"` in my composer.json file, but a correct support on solarium's composer.json would be better :)

thePanz:
Hum, I see what you mean. My PR was against master, thus to early support event-dispatcher 4.x also on solarium 3.x releases

mkalkbrenner:
I don't think that we do another 3.x release.
The develop branch is already merged into 4.x.
And 4.x is backward compatible except that the 3.x Suggester is now called Spellcheck in 4.x. The Suggester itself is fixed. We require that API break to be compatible to the the supported Solr 6 and 7 versions.
So I think the next release will be 4.0.0-alpha1.

Regarding this patch here. There're small differences between the current implementation in 4.x. Can yo please review them?

thePanz:
@mkalkbrenner what if the 3.x release would need some security fixes? no more updates there anymore?
IMHO the dependency is just a small improvement to support the new release of such library.

I am also working on the httpplug integration (or refactor), it would result in a BC (as all the curl, guzzle adapters will be removed). It would be better to have such BC in the 4.x, rather than keeping the BC between 4.x and 3.x
What do you think?

mkalkbrenner:
The 3.x branch is more or less unmaintained at the moment.

From my point of view there're already compatibility issues between solarium 3.x and Solr 6.x/7.x.
That's why I offered my help as a co-maintainer but focusing on the 4.x branch targeting Solr 6 and 7.
For me it sounds strange if someone updates to symfony 4 but wants to stay with solarium 3.x and even worse Solr 5, 4, 3, ... ;-)

My plan is to keep BC breaks at a minimum, but if they are required, we should do it. (And getting rid off these dedicated http adapters is definitely worth a BC break. I'm really looking forward to get your PR!)

I won't be able to take the responsibility for an 3.x release which I don't use by myself. In addition the tests are broken and there're unmerged changes in the 3.x develop branch.
The 4.x branch already has more accepted PRs, the develop branch is merged, the tests are working, the test coverage of unit tests is increased and I added real integration tests querying real Solr 6 and 7 servers. In addition we have the test coverage and a lot of integration tests in drupal/search_api_solr 8.x-2.x which already uses solarium 4.x-dev.

Anyway, I can offer to merge your PR in the 3.x develop branch. But I don't see that I'll find the time to support a 3.x release.

thePanz:
@mkalkbrenner thanks for your reply! :)
I can offer to help maintainign the 3.x branch (with Solr[5,4,3]), as I have a project currently using such combination.
If 4.x has working tests and the CI is good, I'd like to have a look and have those tests cover the 3.x too :)

My PR is still in progress, if you agree with my co-maintaning offer, I can push my changes as a branch here and we could work together on it.

ps: great that you're working with Drupal modules! I used to be the maintainer of serch_api_et for D7 (plus some contributions to the Drupal word) ☯️

thePanz:
Closing MR, as discussed separately.
Thanks @mkalkbrenner !

PR #551, opened 2018-01-19
3.x fix travis

mkalkbrenner:


.gitignore
```diff
@@ -4,3 +4,4 @@
 /composer.lock
 /vendor
 /phar/solarium.phar
+.idea
```
thePanz: @mkalkbrenner this should be (ideally) defined as a common ignore pattern on developer's side ... WDYT?

PR #562, opened 2018-01-24
Fix: Remove invalid configuration from phpunit.xml

localheinz:
This PR

* [x] references `phpunit.xsd` in `phpunit.xml`
* [x] removes invalid configuration items

### Before

screen shot 2018-01-24 at 21 51 37

### After

screen shot 2018-01-24 at 21 51 54

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


thePanz:
Can you apply the same fixes to `phpunit.xml.travis`?
btw: is there a way to keep only one phpunit-xml and have different parameters (like coverage) applied only during the travis-ci run?

What I noticed: I run phpunit locally and no errors were shown, but on travis (due to the different config) some `@covers` failed and the PR was marked as failing.

localheinz:
Thank you, @mkalkbrenner and @thePanz!

PR #565, opened 2018-01-25
Escape the phrases in a range query.

nubs:
In order to prevent query injection, the helper methods should escape the phrases.  Otherwise you could end up with something like rangeQuery('1', '2] someOtherField:[* TO *') and leak information, etc.

See #273 for prior discussion on an old branch.

### DIFF
--- a/tests/Core/Query/HelperTest.php
+++ b/tests/Core/Query/HelperTest.php
@@ -28,12 +28,12 @@ class HelperTest extends TestCase
     public function testRangeQueryInclusive()
     {
         $this->assertEquals(
-            'field:[1 TO 2]',
+            'field:["1" TO "2"]',
             $this->helper->rangeQuery('field', 1, 2)
         );
 
         $this->assertSame(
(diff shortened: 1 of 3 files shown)


nubs:
Updated with a comment in changelog

tests/Core/Query/HelperTest.php
```diff
@@ -28,51 +28,51 @@ public function setUp()
...
     {
         $this->assertSame(
-            'field:{1 TO *}',
+            'field:{"1" TO "*"}',
             $this->helper->rangeQuery('field', 1, null, false)
         );
 
         $this->assertSame(
-            'store:{* TO 46,-93}',
+            'store:{"*" TO "46,-93"}',
```
thePanz: Hum: are we sure those are the expected resuts? the second argument `to "46,-93"]` would match a string, 
If the field is an integer, I guess it would not match any result. WDYT @mkalkbrenner?

Do we have any integration test for this scenario?

PR #570, opened 2018-01-26
Cache solr downloads

thePanz:


.travis.yml
```diff
@@ -17,11 +17,13 @@ env:
...
   directories:
     - $HOME/.composer/cache
+    - solr_downloads
 
 before_install:
   - composer global require "hirak/prestissimo:^0.3"
-  - curl -O http://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz
-  - tar -xzf solr-${SOLR_VERSION}.tgz
+  - mkdir -p solr_downloads && ls solr_downloads/
+  - wget -N --continue -nv --tries=3 --directory-prefix=solr_downloads "http://archive.apache.org/dist/lucene/solr/${SOLR_VERSION}/solr-${SOLR_VERSION}.tgz"
```
mkalkbrenner: Does the '-N' option require that archive.apache.org to be online for checking the timestamp?
My previous approach was to check the cache for existence. I think the timestamp of a pined version should not change and we could avoid requests to archive.apache.org at all.

thePanz: yes, it is required. The point here is to avoid having partially-downloaded files.
But I guess the `-nc` parameter could be ok, too

PR #581, opened 2018-03-13
Stream expression support

mkalkbrenner:


src/QueryType/Stream/Expression.php
```diff
@@ -0,0 +1,20 @@
...
+ **/
+class Expression
+{
+    /**
+     * @param string $name
+     * @param array  $arguments
+     *
+     * @return string
+     */
+    public function __call(string $name, array $arguments)
```
thePanz: Isn't this hiding the possibility to specify not-existing expressions on Solr?

mkalkbrenner: Yes. You'll get an exception within the response.
But this approach will cause solarium to work with the latest version of the streaming expressions.
And it will be very hard to provide specific implementations because there are simply too much already described in the official docs:
    cartesianProduct
    classify
    commit
    complement
    daemon
    eval
    executor
    fetch
    having
    leftOuterJoin
    hashJoin
    innerJoin
    intersect
    merge
    null
    outerHashJoin
    parallel
    priority
    reduce
    rollup
    scoreNodes
    select
    sort
    top
    unique
    update
    search
    jdbc
    echo
    facet
    features
    nodes
    knn
    model
    random
    significantTerms
    shortestPath
    shuffle
    stats
    timeseries
    train
    topic
    tuple
    abs
    acos
    add
    analyze
    and
    anova
    array
    asin
    atan
    betaDistribution
    binomialCoefficient
    binomialDistribution
    cbrt
    ceil
    col
    constantDistribution
    conv
    copyOf
    copyOfRange
    corr
    cos
    cosineSimilarity
    cov
    cumulativeProbability
    derivative
    describe
    diff
    distance
    div
    dotProduct
    ebeAdd
    ebeDivide
    ebeMultiple
    ebeSubtract
    empiricalDistribution
    enumeratedDistribution
    eor
    eq
    expMovingAge
    factorial
    finddelay
    floor
    freqTable
    geometricDistribution
    gammaDistribution
    grandSum
    gt
    gteq
    hist
    hsin
    if
    length
    loess
    log
    logNormalDistribution
    kolmogorovSmirnov
    lt
    lteq
    markovChain
    matrix
    meanDifference
    minMaxScale
    mod
    monteCarlo
    movingAvg
    movingMedian
    mult
    normalDistribution
    normalizeSum
    not
    olsRegress
    or
    poissonDistribution
    polyFit
    pow
    predict
    primes
    probability
    rank
    raw
    regress
    rev
    round
    sample
    scalarAdd
    scalarDivide
    scalarMultiply
    scalarSubtract
    scale
    sequence
    sin
    spline
    sqrt
    standardize
    sub
    sumDifference
    sumColumns
    sumRows
    transpose
    triangularDistribution
    uniformDistribution
    uniformIntegerDistribution
    unitize
    weibullDistribution
    zipFDistribution

And there're already more like "concat" that are just documented in the java docs yet.

PR #586, opened 2018-04-04
Added support for "sow" parameter (Split On Whitespace)

thomascorthals:
Using &sow=false is possible since Solr 6.5. It's needed when you're using multi-word synonyms with SynonymGraphFilter.


src/Component/EdisMax.php
```diff
@@ -200,4 +200,31 @@ public function getUserFields()
...
+
+    /**
+     * Get SplitOnWhitespace option.
+     *
+     * @return bool|null
+     */
+    public function getSplitOnWhitespace()
+    {
+        return $this->getOption('splitonwhitespace');
+    }
```
mkalkbrenner: According to https://lucidworks.com/2017/04/18/multi-word-synonyms-solr-adds-query-time-support/ the `sow`parameter is available for the `edismax` and the `lucene` query parsers.

Depending on your solrconf.xml, you use the lucene query parser if you don't add the EdisMax component to Solarium\QueryType\Select\Query

Therefore I suggest to move this functionality to Solarium\QueryType\Select\Query itself. I assume that the sow parameter simply doesn't hurt if someone sets it in combination with dismax for example.

src/Component/RequestBuilder/EdisMax.php
```diff
@@ -46,6 +46,7 @@ public function buildComponent($component, $request)
         $request->addParam('bf', $component->getBoostFunctions());
         $request->addParam('boost', $component->getBoostFunctionsMult());
         $request->addParam('uf', $component->getUserFields());
+        $request->addParam('sow', null === ($sow = $component->getSplitOnWhitespace()) ? null : ($sow ? 'true' : 'false'));
```
mkalkbrenner: The special handling of null and the string representation of booleans is not required here. We already do that internally in addParam().

tests/Component/RequestBuilder/EDisMaxTest.php
```diff
@@ -31,6 +31,7 @@ public function testBuildComponent()
         $component->setBoostFunctions('functionX(price)');
         $component->setBoostFunctionsMult('functionX(date)');
         $component->setUserFields('title *_s');
+        $component->setSplitOnWhitespace(false);
```
mkalkbrenner: Same as above. The test should be moved to Solarium\Tests\QueryType\Select\RequestBuilderTest

src/Component/RequestBuilder/EdisMax.php
```diff
@@ -46,6 +46,7 @@ public function buildComponent($component, $request)
...
 
         return $request;
     }
```
mkalkbrenner: The new parameter should at least be described within docs/queries/select-query/building-a-select-query/building-a-select-query.md

thomascorthals:
I have moved this functionality to Solarium\QueryType\Select\Query and updated the tests.

PR #591, opened 2018-05-04
Fix return type Result::getFacetSet()

thePanz:


thePanz:
Thanks @mkalkbrenner :+1:

PR #592, opened 2018-05-04
Fix return type Result::getFacetSet() (#591)

mkalkbrenner:
Fix return type Result::getFacetSet()


src/QueryType/Select/Result/Result.php
```diff
@@ -3,8 +3,8 @@
 namespace Solarium\QueryType\Select\Result;
 
 use Solarium\Component\ComponentAwareQueryInterface;
+use Solarium\Component\FacetSet;
```
mkalkbrenner: Isn't it Solarium\Component\Result\FacetSet ?

PR #597, opened 2018-05-15
 Added Query Elevation Component

thomascorthals:
I've added support for Solr's [Query Elevation Component](https://lucene.apache.org/solr/guide/the-query-elevation-component.html).

The solarium component does three things:

1. Override the request handler (`elevate` by default).
2. Add document transformers to the existing `fl` parameter.
3. Add optional additional parameters to the request.

Test coverage and documentation is included.

### DIFF
--- a/src/Component/QueryElevation.php
+++ b/src/Component/QueryElevation.php
@@ -17,7 +17,6 @@ class QueryElevation extends AbstractComponent
      * @var array
      */
     protected $options = [
-        'handler' => 'elevate',
         'transformers' => '[elevated]',
     ];
 
@@ -53,28 +52,6 @@ class QueryElevation extends AbstractComponent
      */
     public function getResponseParser()
     {
-    }
-
-    /**
-     * Set handler option.
-     *
-     * @param string $handler
-     *
-     * @return self Provides fluent interface
-     */
-    public function setHandler($handler)
-    {
-        return $this->setOption('handler', $handler);
-    }
-
-    /**
-     * Get handler option.
-     *
-     * @return string
-     */
-    public function getHandler()
-    {
-        return $this->getOption('handler');
     }
 
     /**
--- a/src/Component/QueryElevation.php
+++ b/src/Component/QueryElevation.php
@@ -328,7 +328,7 @@ class QueryElevation extends AbstractComponent
         foreach ($this->options as $name => $value) {
             switch ($name) {
                 case 'transformers':
-                    $this->addTransformers($value);
+                    $this->setTransformers($value);
                     break;
                 case 'markExcludes':
                     $this->setMarkExcludes($value);
--- a/src/Component/QueryElevation.php
+++ b/src/Component/QueryElevation.php
@@ -246,8 +246,7 @@ class QueryElevation extends AbstractComponent
     {
         if (true === $mark || 'true' === $mark) {
             $this->addTransformer('[excluded]');
-        }
-        else {
+        } else {
             $this->removeTransformer('[excluded]');
         }
 
--- a/src/Component/QueryElevation.php
+++ b/src/Component/QueryElevation.php
@@ -1,0 +1,346 @@
+<?php
+
+namespace Solarium\Component;
+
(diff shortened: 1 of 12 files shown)


src/Component/QueryElevation.php
```diff
@@ -0,0 +1,345 @@
...
+ */
+class QueryElevation extends AbstractComponent
+{
+    /**
+     * Default options.
+     *
+     * @var array
+     */
+    protected $options = [
+        'handler' => 'elevate',
```
mkalkbrenner: I don't think that this is required. Or it's wrong!

According to the documentation the dedicated elevate handler exists in the techproducts example only "for brevity". But the component "may be added to any request handler".

src/Component/QueryElevation.php
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    /**
+     * Set handler option.
+     *
+     * @param string $handler
+     *
+     * @return self Provides fluent interface
+     */
+    public function setHandler($handler)
```
mkalkbrenner: Might be not required.

src/Component/QueryElevation.php
```diff
@@ -0,0 +1,345 @@
...
+    {
+        return $this->setOption('handler', $handler);
+    }
+
+    /**
+     * Get handler option.
+     *
+     * @return string
+     */
+    public function getHandler()
```
mkalkbrenner: Might be not required.

src/Component/RequestBuilder/QueryElevation.php
```diff
@@ -0,0 +1,45 @@
...
+     *
+     * @param QueryelevationComponent $component
+     * @param Request                 $request
+     *
+     * @return Request
+     */
+    public function buildComponent($component, $request)
+    {
+        // switch request handler
+        $request->setHandler($component->getHandler());
```
mkalkbrenner: A component should never secretly switch the request handler. This might break the main query.

tests/Component/QueryElevationTest.php
```diff
@@ -0,0 +1,202 @@
...
+
+    public function setUp()
+    {
+        $this->queryelevation = new QueryElevation();
+    }
+
+    public function testConfigMode()
+    {
+        $options = [
+            'handler' => 'myhandler',
```
mkalkbrenner: Should be removed.

tests/Component/QueryElevationTest.php
```diff
@@ -0,0 +1,202 @@
...
+            'forceElevation' => true,
+            'exclusive' => true,
+            'markExcludes' => false,
+            'elevateIds' => 'doc1,doc2',
+            'excludeIds' => 'doc3,doc4',
+        ];
+
+        $this->queryelevation->setOptions($options);
+
+        $this->assertSame('myhandler', $this->queryelevation->getHandler());
```
mkalkbrenner: Should be removed.

tests/Component/QueryElevationTest.php
```diff
@@ -0,0 +1,202 @@
...
+
+    public function testGetRequestBuilder()
+    {
+        $this->assertInstanceOf(
+            'Solarium\Component\RequestBuilder\QueryElevation',
+            $this->queryelevation->getRequestBuilder()
+        );
+    }
+
+    public function testSetAndGetHandler()
```
mkalkbrenner: Should be removed.

src/Component/QueryElevation.php
thomascorthals: If you want to use Query Elevation on `/select`, you have to add it in `solrconfig.xml`. I (wrongfully?) assumed `/elevate` was always included and therefore it would "just work" this way even if you don't have access rights to edit the config.

I'll remove it and include a note about the config in the documentation. If you use the Solarium component on `/select` without adding the Solr component, all it does is cause head-scratching.

thomascorthals:
I've handled the handler situation and added a test for both elevation and exclusion against techproducts.

PR #614, opened 2018-08-08
[BUGFIX] Extract request rejected because of missing multipart boundary

timohund:
When a ExtractQuery is executed Apache Solr throws the following error:

* SEVERE: null:org.apache.commons.fileupload.FileUploadException: the request was rejected because no multipart boundary was found

The patch:

* Add's the boundary to the header and to the body
* Add's integration tests for the extract query

Fixes: #301


mkalkbrenner:
@timohund thanks for your work! It's good to have you participating now :-)

@thePanz Thanks for reviewing the patch.

timohund:
What is a little bit strage is that coverall indicates a decrease of the coverage, to should normally not be the case since all parts are covered at least by the integration tests.

Issue #618, opened 2018-08-08
DateTime is not set to UTC when using Extract


jeroenherczeg:
When using the update QueryType, it will change the timezone to UTC.
When using the extract QueryType, it will keep **not** change it to UTC.

jeroenherczeg:
Would it be possible to get this fix in v3?

mkalkbrenner:
Sure, if @thePanz agrees and if you file a new PR for that, or at least a new issue.

jeroenherczeg:
I will create a new PR, which branch should it be based of?

mkalkbrenner:
3.x
https://github.com/solariumphp/solarium/tree/3.x

PR #619, opened 2018-08-08
Fix for #618 DateTime is not set to UTC when using Extract 

jeroenherczeg:
Fix for https://github.com/solariumphp/solarium/issues/618


jeroenherczeg:
Thank you!

PR #622, opened 2018-08-12
Update FacetSet.php

je4:
when using multiple facets on the same field, the facet parameter (prefix, limit etc.) are connected to the field instead of the facet key. therefor, they are valid for all facets on one field.

current solr versions allow local params to the facet.field parameter. therefor, as much parameter as possible should be moved to the facet.field.

i am not sure, whether this can be used directly or whether there should be a switch for that.
apparently, since SOLR 6.6 (eventually earlier), local parameters are possible for the facet.field component.

```
$facetSetCategory = $squery->getFacetSet();
$facetSetCategory->createFacetField('category')->setField('category')->setLimit( 50 )->addExclude('category');
$facetSetArea = $squery->getFacetSet();
$facetSetArea->createFacetField('area')->setField('category')->setPrefix( '1!!area' )->setLimit( 10 )->addExclude('area');
```
thank you
jürgen

### DIFF
--- a/tests/Component/RequestBuilder/FacetSetTest.php
+++ b/tests/Component/RequestBuilder/FacetSetTest.php
@@ -303,7 +303,7 @@ class FacetSetTest extends TestCase
     {
         $this->component->setMissing(true);
         $this->component->setLimit(10);
-        $this->component->addFacet(new FacetField(['key' => 'f1', 'field' => 'owner']))->setPrefix('Y');
+        $this->component->addFacet((new FacetField(['key' => 'f1', 'field' => 'owner']))->setPrefix('Y'));
 
         // second use of field owner (with prefix)
         $this->component->addFacet((new FacetField(['key' => 'f2', 'field' => 'owner']))->setPrefix('X'));


src/Component/RequestBuilder/FacetSet.php
```diff
@@ -117,19 +117,30 @@ public function addFacetField($request, $facet)
             'facet.field',
             $this->renderLocalParams(
                 $field,
-                ['key' => $facet->getKey(), 'ex' => $facet->getExcludes()]
+                ['key' => $facet->getKey()
+                  , 'ex' => $facet->getExcludes()
```
mkalkbrenner: These lines violate the coding standard, see https://github.styleci.io/analyses/q5l0Ey?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

src/Component/RequestBuilder/FacetSet.php
```diff
@@ -117,19 +117,30 @@ public function addFacetField($request, $facet)
...
-        $request->addParam("f.$field.facet.limit", $facet->getLimit());
-        $request->addParam("f.$field.facet.sort", $facet->getSort());
-        $request->addParam("f.$field.facet.prefix", $facet->getPrefix());
-        $request->addParam("f.$field.facet.contains", $facet->getContains());
-        $request->addParam("f.$field.facet.contains.ignoreCase", $facet->getContainsIgnoreCase());
-        $request->addParam("f.$field.facet.offset", $facet->getOffset());
-        $request->addParam("f.$field.facet.mincount", $facet->getMinCount());
-        $request->addParam("f.$field.facet.missing", $facet->getMissing());
-        $request->addParam("f.$field.facet.method", $facet->getMethod());
+        //$request->addParam("f.$field.facet.limit", $facet->getLimit());
```
mkalkbrenner: We should entirely remove these lines. Do we know which minimum Solr version is required to switch to local params?

mkalkbrenner: We have failing tests on travis:
```
1) Solarium\Tests\Component\RequestBuilder\FacetSetTest::testBuildWithContainsSettings
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'?facet.field={!key=f1}owner&f.owner.facet.contains=foo&f.owner.facet.contains.ignoreCase=true&facet=true&facet.contains=bar&facet.contains.ignoreCase=false'
+'?facet.field={!key=f1 facet.contains=foo facet.contains.ignoreCase=1}owner&facet=true&facet.contains=bar&facet.contains.ignoreCase=false'
/home/travis/build/solariumphp/solarium/tests/Component/RequestBuilder/FacetSetTest.php:378
```

At least the failing test needs to be adjusted. But is the new output correct from your perspective?
```
?facet.field={!key=f1 facet.contains=foo facet.contains.ignoreCase=1}owner&facet=true&facet.contains=bar&facet.contains.ignoreCase=false
```

je4:
There are more candidates which are could use local parameters:
* public function addFacetRange($request, $facet)
* public function addFacetPivot($request, $facet)
* public function addFacetInterval($request, $facet)
but i do not have a test environment to figure out, whether it works.

mkalkbrenner:
I wonder if https://issues.apache.org/jira/browse/SOLR-6193 is till an issue.
Anyway, It might make sense to keep the old behavior for backward compatibility and to work around the Solr issue for distributed search users.

What do you think about dynamically switching to local parameters as soon as a field is used for facetting more then once? That should be easy to implement within that single file.

je4:
i made a change, which makes sure, that the local parameter are only used if one field appears multiple times. it's quite short and hopefully usable.

je4:
according to https://issues.apache.org/jira/browse/SOLR-6193 
this should be no problem, because the parameters are local to facet.field OR external. therefor the external one, which would override the local one, does not exists.

src/Component/RequestBuilder/FacetSet.php
```diff
@@ -108,28 +118,31 @@ public function buildComponent($component, $request)
      *
      * @param Request    $request
      * @param FacetField $facet
+     * @param compatible $compatible false if local parameters should be used
      */
-    public function addFacetField($request, $facet)
+    public function addFacetField($request, $facet, $compatibility_mode = true)
```
mkalkbrenner: I prefer to name the variable $use_local_params and to invert the logic. So the default would be false.

src/Component/RequestBuilder/FacetSet.php
```diff
@@ -108,28 +118,31 @@ public function buildComponent($component, $request)
      *
      * @param Request    $request
      * @param FacetField $facet
+     * @param compatible $compatible false if local parameters should be used
```
mkalkbrenner: The comment has to be enhanced to explain when local params make sense, just a short sentence. And we should mention https://issues.apache.org/jira/browse/SOLR-6193 to explain why local params aren't always used.

src/Component/RequestBuilder/FacetSet.php
```diff
@@ -108,28 +118,31 @@ public function buildComponent($component, $request)
...
      */
-    public function addFacetField($request, $facet)
+    public function addFacetField($request, $facet, $compatibility_mode = true)
     {
         $field = $facet->getField();
 
+        $local_param = $compatibility_mode ? ['key' => $facet->getKey(), 'ex' => $facet->getExcludes()] : ['key' => $facet->getKey(), 'ex' => $facet->getExcludes(), 'facet.limit' => $facet->getLimit(), 'facet.sort' => $facet->getSort(), 'facet.prefix' => $facet->getPrefix(), 'facet.contains' => $facet->getContains(), 'facet.contains.ignoreCase' => $facet->getContainsIgnoreCase(), 'facet.offset' => $facet->getOffset(), 'facet.mincount' => $facet->getMinCount(), 'facet.missing' => $facet->getMissing(), 'facet.method' => $facet->getMethod()];
```
mkalkbrenner: The two long arrays should be formatted to set each param in a separate line. In your previous code your only mistake according to the style checker was that you didn'tplace the commas at the end of the line.

src/Component/RequestBuilder/FacetSet.php
```diff
@@ -108,28 +118,31 @@ public function buildComponent($component, $request)
...
         $request->addParam(
             'facet.field',
             $this->renderLocalParams(
                 $field,
-                ['key' => $facet->getKey(), 'ex' => $facet->getExcludes()]
+                $local_param
```
mkalkbrenner: The variable should be named plural: $local_params.

je4:
i think, we can start merging now...

je4:
how about the other candidates?
* public function addFacetRange($request, $facet)
* public function addFacetPivot($request, $facet)
* public function addFacetInterval($request, $facet)
do you think, it's the same problem? i don't have an easy test for these facet types in at my system...

tests/Component/RequestBuilder/FacetSetTest.php
```diff
@@ -297,7 +299,7 @@ public function testBuildWithFacetsAndGlobalFacetSettings()
         );
 
         static::assertEquals(
-            '?facet.field={!key=f1}owner&facet.query={!key=f2}category:23&facet.query={!key=f4}category:40&facet=true&facet.missing=true&facet.limit=10',
```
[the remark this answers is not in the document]
je4: just added a new assert to the old code...
hope that works...

mkalkbrenner:
It's good to have tests ;-)

```
1) Solarium\Tests\Component\RequestBuilder\FacetSetTest::testBuildWithFacetsAndSameFieldMultiplePrefix
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'?facet.field={!key=f1 facet.prefix=Y}owner&facet.field={!key=f2 facet.prefix=X}owner&facet=true&facet.missing=true&facet.limit=10'
+'?facet.field={!key=f1}owner&facet.field={!key=f2 facet.prefix=X}owner&facet=true&facet.prefix=Y&facet.missing=true&facet.limit=10'
/home/travis/build/solariumphp/solarium/tests/Component/RequestBuilder/FacetSetTest.php:320
```
I think the expected result is correct but either your new code or the existing renderLocalParams() function must be erroneous.

tests/Component/RequestBuilder/FacetSetTest.php
```diff
@@ -288,6 +288,25 @@ public function testBuildWithFacetsAndGlobalFacetSettings()
...
+            '?facet.field={!key=f1}owner&facet.query={!key=f2}category:23&facet.query={!key=f4}category:40&facet=true&facet.missing=true&facet.limit=10',
+            urldecode($request->getUri())
+        );
+    }
+
+    public function testBuildWithFacetsAndSameFieldMultiplePrefix()
+    {
+        $this->component->setMissing(true);
+        $this->component->setLimit(10);
+        $this->component->addFacet(new FacetField(['key' => 'f1', 'field' => 'owner']))->setPrefix('Y');
```
mkalkbrenner: You set the Prefix on the FacetSet. Wouldn't it be correct to set the Prefix on the Facet?
```
$this->component->addFacet(new FacetField(['key' => 'f1', 'field' => 'owner'])->setPrefix('Y'));
```

je4: Ups, wrong bracket counting...

mkalkbrenner:
>how about the other candidates?
>
>    public function addFacetRange($request, $facet)
>    public function addFacetPivot($request, $facet)
>    public function addFacetInterval($request, $facet)
>
>do you think, it's the same problem? i don't have an easy test for these facet types in my system...

I suggest to open a new PR for these and to merge the current one first.
Why? The current patch is backward compatible and therefore not dangerous.
But before we extend it to features we're not experienced with, we should add some other tests. Our test suite contains some real integration tests against a real Solr index. We should extend them to include the facet types in question. These tests will ensure that we don't break anything when we touch the parameters.

PR #623, opened 2018-08-14
Solarium 3.x: Fix for #618 DateTime is not set to UTC when using Extract in v3.x

jeroenherczeg:
Fix for https://github.com/solariumphp/solarium/issues/618

### DIFF
--- a/tests/Solarium/Tests/QueryType/Extract/RequestBuilderTest.php
+++ b/tests/Solarium/Tests/QueryType/Extract/RequestBuilderTest.php
@@ -139,4 +139,26 @@ class RequestBuilderTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($headers,
                             $request->getHeaders());
     }
+
+    public function testDocumentDateTimeField()
+    {
+        $timezone = new \DateTimeZone('Europe/London');
+        $date = new \DateTime('2013-01-15 14:41:58', $timezone);
+        $document = $this->query->createDocument(array('date' => $date));
+        $this->query->setDocument($document);
+        $request = $this->builder->build($this->query);
+        $this->assertEquals(
+            array(
+                'fmap.from-field' => 'to-field',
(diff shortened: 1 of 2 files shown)


jeroenherczeg:
Should I worry about styleci?

tests/Solarium/Tests/QueryType/Extract/RequestBuilderTest.php
```diff
@@ -139,4 +139,26 @@ public function testContentTypeHeader()
         $this->assertEquals($headers,
                             $request->getHeaders());
     }
+
+    public function testDocumentDateTimeField()
+    {
+        $timezone = new \DateTimeZone('Europe/London');
+        $date = new \DateTime('2013-01-15 14:41:58', $timezone);
+        $document = $this->query->createDocument(array('date' => $date));
```
thePanz: Aren't we using the short-form for arrays? (I thought it was covered by the php-cs-styles) :/

jeroenherczeg:
In 3x branch, composer.json says this package supports PHP 5.3.9, I think short form array syntax was introduced in 5.4

thePanz:
@jeroenherczeg true. I am quite scared about this library still supporting 5.3 (see: http://php.net/supported-versions.php). should we move to 5.6 (and deprecate this library's v3 after php 5.6 is out of life? wdyt?)

jeroenherczeg:
Moving to 5.6 for v3 would be a breaking change not in a major version, could introduce a problem for someone, but then again still running 5.3 is also a problem. 

Deprecating v3 at the same time with 5.6, sounds good to me.

jeroenherczeg:
Is it possible to merge this PR, or does it require changes?

Schnaaf:
Can this be reviewed please?

jeroenherczeg:
@thePanz Is there something blocking this PR?

mkalkbrenner:

Issue #624, opened 2018-08-17
Support core admin api


timohund:
As discussed in #490 it would be nice to support the core admin api (https://lucene.apache.org/solr/guide/6_6/coreadmin-api.html). I'll get this started in the next time.

PR #625, opened 2018-08-20
[TASK] Implement core admin api for solarium

timohund:
This pr:

* Implements the CoreAdmin, which can be executed with several actions.

Example:

* Add's unit tests for the implemented CoreAdmin actions
* Refactores the Request and Query to indicated if a query is coreIndependent. Which means that this query needs to be executed outside of a core context.

Resolves: #624


timohund:
Just as base for discussion needs some changes.

timohund:
Hi,

@mkalkbrenner @thePanz  @basdenooijer  i've pushed a first version of the core admin api.

By now it is backwards compatible and i tried to implement it in a way that it fit's into the current structure of solarium. It would be nice to get your Feedback.

There are something where it would be good that we discuss them:

* By now all Querys are in the QueryType package. Do you think it makes sence to split them into Admin in NonAdmin packages to have them logically separated? As it is implemented now, it is not seperated because i think this requires some breaking changes

* The implementation of the core admin feature requires to distinguish between coreindependent and corerelated query types. By now every query was corerelated, which means that this Query allways needs to be executed in the context of a core/collection. This is now solved by passing a boolean value to the getBaseUri method. This is backwards compatible but i think it would be better to have two methods getBaseUri and getCoreBaseUri. This would require to make a breaking change in getBaseUri(). 

What do you think? By now everything is implemented to keep backwards compatibility but it might makes sence to introduce so breaking changes in version 5.

src/QueryType/CoreAdmin/Query/Action/AbstractAction.php
```diff
@@ -0,0 +1,18 @@
...
+ * CoreAdmin query command base class.
+ */
+abstract class AbstractAction extends Configurable
+{
+    /**
+     * Returns command type, for use in adapters.
+     *
+     * @return string
+     */
+    abstract public function getType();
```
thePanz: you added the typehint for the other methods. the `getType(): string;` is missing here

src/Core/Query/QueryInterface.php
```diff
@@ -94,4 +94,11 @@ public function addParam($name, $value);
      * @return array
      */
     public function getParams();
+
+    /**
+     * Indicates if a query is core independent and should be executed outside of a core context.
+     *
+     * @return bool
+     */
+    public function getIsCoreIndependent();
```
thePanz: Missing return typehint

timohund: thx, fixed that

src/QueryType/CoreAdmin/Query/Action/AbstractAction.php
timohund: thx, fixed that

mkalkbrenner:
First of all, great work so far!

I'm interested in getting this in very quickly.

> By now all Querys are in the QueryType package. Do you think it makes sence to split them into Admin in NonAdmin packages to have them logically separated? As it is implemented now, it is not seperated because i think this requires some breaking changes

In which way will this cause breaking changes?

> The implementation of the core admin feature requires to distinguish between coreindependent and corerelated query types. By now every query was corerelated, which means that this Query allways needs to be executed in the context of a core/collection. This is now solved by passing a boolean value to the getBaseUri method. This is backwards compatible but i think it would be better to have two methods getBaseUri and getCoreBaseUri. This would require to make a breaking change in getBaseUri().

Even the Name getCoreBaseUri() is wrong in the context of Solr Cloud collections.
I suggest to keep getBaseUri() for the core/collection (transparently, depending if it is a Solr Cloud or not). In addition we should introduce getServerUri() and remove the isCoreIndependent stuff.
(The Drupal modules evolved the same way.)

timohund:
@mkalkbrenner  thanks for your feedback:

> In which way will this cause breaking changes?

Because we would have two new packages below QueryType e.g. Server (for the server related querys like the admin query) and Core (for the core specific query, everything else then the CoreAdmin queries by now). I am not happy with these names by now but i would say we would need one package for server specific queries and one for the core/collection specific query. Or we keep the queries as they are and just add a new package for the Server specific.

Do you have a proposal there?

>> Even the name getCoreBaseUri() is wrong in the context of Solr Cloud collections.
>> I suggest to keep getBaseUri() for the core/collection (transparently, depending if it is a Solr >>Cloud or not). In addition we should introduce getServerUri() and remove the isCoreIndependent >> stuff. (The Drupal modules evolved the same way.)

Having getCoreBaseUri and getServerUri is good. With removing the "isCoreIndependet" stuff, do you mean just to remove the boolean value here? Because on the query/request level we need to have this information to be able to decide if "getCoreBaseUri" or "getServerUri" needs to be called.

As i understand you right i would:

* Remove the option "coreindependent" from the query.
* Move the CoreAdmin package into an own subpackage "QueryType\Server\CoreAdmin"
* Have a property in the request to distinguish between server and non-server requests
* In the Adapters call "getCoreBaseUri" or "getServerUri" depending if we have a server or core/collection specific request.

Is that right? Do you have other proposals?

mkalkbrenner:
> Or we keep the queries as they are and just add a new package for the Server specific.

That's the backward compatible way I would prefer.

getCoreBaseUri and getServerUri will be new functions.
getBaseUri will be declared deprecated and directly call getCoreBaseUri for some releases.

timohund:
@mkalkbrenner just pushed the current state, in the next step i'll add some more examples and complete the tests for some parts. If you have anything else please let me know.

timohund:
@mkalkbrenner @thePanz i've added the feature, documentation and tests can you maybe review it and tell me if there is something left to merge it?

PR #627, opened 2018-08-22
Remove invalid aliases

Seldaek:


thePanz:
Thanks for this PR @Seldaek , I am still confused by the remaining alias (from the docs: https://getcomposer.org/doc/articles/aliases.md#branch-alias).

Shouldn't it be:
```
    "extra": {
        "branch-alias": {
            "dev-master": "dev-4.1.x"
        }
    }
```
The latest minor release of this library is 4.1, thus I guess that the `master` branch should be aliased to `dev-4.1.x`.
We already have a `4.0.x` branch, and it could be selected by using `dev-4.0.x`.

Let me know if it's correct :)

thePanz:
Ok, thanks for the clarification!
So, the following should be the correct one? If yes can you update your PR, so that I can merge it to `master`? Thank you
```
"extra": {
        "branch-alias": {
            "dev-master": "4.1.x-dev"
        }
    }
```

Seldaek:
That would be correct. On my phone now so feel free to do the edit and
close my PR :)

thePanz:
Thanks @Seldaek , fixed! I'll merge

Issue #626, opened 2018-08-22
Support managed resources api


timohund:
Stopword and Synonyms can be manged with managed resources in Apache Solr. It would be nice to support the API for that in solarium. Some years ago i started that in an own package https://github.com/timohund/solr-rest-api-client and it would be nice to have this functionality in solarium.

I'll try to implement that in the next weeks.

jsteggink:
Hi, I just built this. I'll try if I can create a pull request this week.

timohund:
Sounds good!

timohund:
For synonyms solr allows to create multiple entries, without passing a base word (a group with equal words).

E.g. "iphone", "smartphone", "mobile"

=>
"iphone" => "smartphone", "mobile"
"smartphone" => "iphone", "mobile"
"mobile" => "iphone", "smartphone"

I think i would be cool to support that and maybe "detect" on the api level the groups. To group them in the rendering, or to delete them in one api call. What do you think? If i can help you with testing or whatever just ping me.

jsteggink:
@timohund, sorry, I didn't get to it as I'm on a holiday. I'll do a PR on Monday.

jsteggink:
@timohund I pushed a commit, maybe you can have a look: https://github.com/jsteggink/solarium/tree/issue-626-managed-resources

timohund:
@jsteggink 

Thanks for your work, would be nice to have that in Version 5 of solarium.

By reading the code i have a few questions:

1. For the commands there is a method "addCommand". Is it possible to execute multiple commands or just one at once? When only one is possible i would name it "setCommand", for me "addCommand" sounds like you are able to run multiple commands.
2. What about a GET_COMMAND, i guess this is just missing by now?
3. I think i would be nice to have some integration tests that cover the creation, deletion and retrieval of those entities to make sure that this still works when a newer version of Apache Solr will be used.

Maybe @mkalkbrenner @basdenooijer  and @thePanz can also give some input. By now i think this is really a cool feature and what you did looks great.

jsteggink:
Thanks for the review @timohund!

1. I will change it to setCommand
2. Right now it works by just firing the query without any command. I followed the pattern of other query types where it works like this. All the other commands require an additional parameter, however, GET has no parameters. It just lists all resources.
3. I have made the integration tests, but the schema doesn't have the right filters yet. I'm not adding it yet, because I'm unsure if it affects any other tests as of now. I'll have to check.

[...]

PR #632, opened 2018-10-08
Issue #626 managed resources

jsteggink:
Adds new features according issue #626.


jsteggink:
Thanks for your pointers. However, I have already added the integration tests (AbstractTechproductsTest.php, (methods: testManagedStopwords, testManagedSynonyms and testManagedResources). They are based on the default techproducts schema (server/solr/configsets/sample_techproducts_configs/conf) which is included with Solr. This schema includes the managed resources I use in the integration tests.

Or are you missing something else?

mkalkbrenner:
Oh, please excuse. I scrolled to fast and didn't recognize this test :-)
Maybe just one addition to the test:
Check for non existence of the stopword and the synonyms first before adding them.

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -8,6 +8,7 @@
 use Solarium\Component\Result\Terms\Result;
 use Solarium\Core\Client\ClientInterface;
 use Solarium\Exception\HttpException;
+use Solarium\QueryType\ManagedResources\Query\Stopwords\Command\Add;
```
mkalkbrenner: Maybe `as AddStopwords`?
And the other commands are completely missing.

src/QueryType/ManagedResources/RequestBuilder/Synonyms.php
```diff
@@ -0,0 +1,63 @@
...
+
+use RuntimeException;
+use Solarium\Core\Client\Request;
+use Solarium\Core\Query\AbstractRequestBuilder as BaseRequestBuilder;
+use Solarium\Core\Query\QueryInterface;
+use Solarium\QueryType\ManagedResources\Query\Synonyms as SynonymsQuery;
+use Solarium\QueryType\ManagedResources\Query\Synonyms\Command\AbstractCommand;
+
+
+class Synonyms extends BaseRequestBuilder {
```
mkalkbrenner: This request builder is very similar to Stopwords. I would avoid that redundant code using a base class or a trait.

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -647,6 +648,104 @@ public function testGetStatusFromAllCoreWhenNoCoreNameWasSet()
...
+    {
+        $query = $this->client->createManagedStopwords();
+        $query->setName('english');
+        $term = 'the';
+
+        // Add stopwords
+        $add = new AddStopwords();
+        $add->setStopwords([$term]);
+        $query->setCommand($add);
+        $result = $this->client->execute($query);
```
mkalkbrenner: According to https://lucene.apache.org/solr/guide/7_5/managed-resources.html#managed-resources "you must reload the core/collection before changes become active".

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -647,6 +648,104 @@ public function testGetStatusFromAllCoreWhenNoCoreNameWasSet()
...
+
+        // Check if single stopword exist
+        $exists = new ExistsStopwords();
+        $exists->setTerm($term);
+        $query->setCommand($exists);
+        $result = $this->client->execute($query);
+        $this->assertEquals(200, $result->getResponse()->getStatusCode());
+
+        // List stopwords
+        $query->removeCommand();
```
mkalkbrenner: move this line above the comment and add another comment to explain the purpose, even if I already got it ;-)

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -647,6 +648,104 @@ public function testGetStatusFromAllCoreWhenNoCoreNameWasSet()
...
+        // Check if single synonym exist
+        $exists = new ExistsSynonyms();
+        $exists->setTerm($term);
+        $query->setCommand($exists);
+        $result = $this->client->execute($query);
+        $this->assertEquals(200, $result->getResponse()->getStatusCode());
+        $this->assertSame(['mad' => ['angry', 'upset']], $result->getData());
+
+        // List synonyms
+        $query->removeCommand();
```
mkalkbrenner: Again, move this line.

tests/Integration/AbstractTechproductsTest.php
jsteggink: Reloading is only necessary for Solr components (e.g. ManagedSynonymGraphFilterFactory). It has nothing to with our tests. We can't just blindly reload the core/collection because it has all sorts of implications. It would be an action purposely initiated by a user.

mkalkbrenner: I agree, that we should not trigger a reload after the command automatically. But it needs to be documented that the user needs to reload the core or the collection at some point.

I just wondered if the tests will fail due to that. Right now they didn't run because of these class not found issues.

mkalkbrenner:
Tests passed! I'll have a look ...

jsteggink:
What happened? The pull request is closed, but the code wasn't merged to master.

mkalkbrenner:
WTF? I'm sure I hit the merge button.

mkalkbrenner:
And I mentioned it in the CCHANGELOG

PR #657, opened 2019-03-17
Solarium 3.x: ci failure fixes

wickedOne:
this should resolve both failing travis and style ci tests for the 3.x branch

did exclude the ``tests`` directory from php-cs-fxier as it continues to modify instances of ``assertNotEquals`` to ``assertNotFalse`` regardless of the ``php_unit_strict`` settings i try.

fixed hhvm tot v3.3 in travis because: https://hhvm.com/blog/2018/09/12/end-of-php-support-future-of-hack.html

and really tried to get a working php 5.3 environment, but you appaer to have an impossible set of dependencies (guzzle v.s. event dispatcher v.s. coveralls).

i think the main discussion should be how long you're willing to continue supporting php 5.3

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,37 @@
+<?php
+
+return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
+    ->setRules([
+       '@PHP56Migration' => true,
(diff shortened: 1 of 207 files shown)


mkalkbrenner:

PR #659, opened 2019-03-26
Adjust example 1.1 using composer

clemens-tolboom:
Following comment in https://github.com/solariumphp/solarium/pull/323#issuecomment-411822212

I also assume running from the command line as that clean up the example more. Hope that's fine too.

### DIFF
--- a/examples/1.1-check-solarium-and-ping.php
+++ b/examples/1.1-check-solarium-and-ping.php
@@ -1,10 +1,11 @@
 <?php
 
-require(__DIR__.'/init.php');
-htmlHeader();
+require(__DIR__ . '/vendor/autoload.php');
 
 // check solarium version available
-echo 'Solarium library version: ' . Solarium\Client::VERSION . ' - ';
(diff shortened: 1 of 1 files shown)


examples/1.1-check-solarium-and-ping.php
```diff
@@ -1,10 +1,11 @@
 <?php
 
-require(__DIR__.'/init.php');
-htmlHeader();
+require(__DIR__ . '/vendor/autoload.php');
```
clemens-tolboom: As this is an example maybe better not uss `__DIR__` but

```php
require('./vendor/autoload.php');
```
then people can copy/paste and maybe run immediately.

clemens-tolboom:
Hmmm ... guess I'm wrong in 'No HTML' as 1.3 is harder to convert. Guess it's better to close this and start over making a new PR. Similar for #660.

Keeping this open until feedback.

clemens-tolboom:
This was a bad call

PR #660, opened 2019-03-26
Update 1.2-basic-select using composer

clemens-tolboom:
Similar to #659:

- assume command line so no HTML
- use current directory (which works better for converting other examples using symlinks)


clemens-tolboom:
This was a bad call.

Issue #490, opened 2017-03-08
Solarium 5.0 (what we didn't achieve with 4.x)


jsteggink:
In order to get things going and plan for Solarium 4.0, I'd like to get an idea of what improvements and new features people are looking for.

# Features
- Use the new v2 api
- Query Re-Ranking
- Streaming Expressions
- Admin query types
- Filling feature gaps in the existing QueryTypes

# Improvements
- In #489 we discussed going to PHP 7.
- Splitting up some of the methods to traits in order to make room for SolrCloud. For example the distinction between cores and collections.
- Passing Endpoint objects to methods could be removed. Having a central registration of an Endpoint should be ok.
- Move parameter and requesthandler constants to a separate classes (Params classes)

thePanz:
May I add: "use of PHP Http Plug" library? (http://docs.php-http.org/en/latest/) instead of writing our adapters for cUrl/Guzzle , that library can be used to abstract them (ps: I wrote an API client for mattermost, it is quite easy to use phphttp plug! :) )

mkalkbrenner:
As I already told @basdenooijer, I propose to drop support for all software that reached EOL.
According to http://lucene.apache.org/solr/community.html we should only support Solr 6.4 and 7.x.
But if we take into account that the 6.4 "LTS" didn't receive the critical security issues but 6.6 did, I would suggest that solarium 4 should support Solr > 6.6.x.

According to http://php.net/supported-versions.php PHP 5.6 will be supported till end of 2018. Nevertheless I vote for PHP 7 only.

People who need to run an older PHP version or Solr Server could perfectly stay with an old 3.x solarium release.

mkalkbrenner:
@thePanz I like the idea to get rid of all these specific client implementations and to use an abstraction. Are you able to start such an an PR against the 4.x branch, maybe as WIP?

thePanz:
@mkalkbrenner I can start a first implementation, sure! :)

thePanz:
@mkalkbrenner just a small point, are we moving to PHP 7.1, right?

mkalkbrenner:
@thePanz will PHP 7.1 be required? Or does it work PHP 7.0 as well?

thePanz:
@mkalkbrenner I was asking just to know if I can use PHP 7.1 features or not :)

mkalkbrenner:
@thePanz If it's doable I would prefer an implementation that is compatible to 7.0.

[...]

PR #661, opened 2019-03-27
Use vendor/autoload.php.

clemens-tolboom:
If we extends the config to contain the autoloader reference we can let people run the examples on any configuration as long as the config is in the webserver document root.

This works as described in the updated README.

This replaces PRs #659 and #660.


clemens-tolboom:
I build this patch on v4.2.0 as master (I thought) caused errors on my `Solr Implementation Version: 3.6.2 debian - pbuilder - 2018-05-02 22:38:56`

> Solarium\Exception\UnexpectedValueException: No collection set.

Is that part of the discussion in #490 

> Splitting up some of the methods to traits in order to make room for SolrCloud. For example the distinction between cores and collections. => **partly done**

mkalkbrenner:
Since I never used this config feature, I need to take a closer look first. But configuring the autoloadr this way in a composer managed project seems suspicious.

PR #682, opened 2019-07-09
Sf4.3 testing and fix deprecations

DemigodCode:
Hi,
I've added symfony 4.3 as testing cases and I've fixed the deprecations thrown cause of the event dispatcher.


DemigodCode:
Hey @mkalkbrenner is there an opportunity to merge this pr?

src/Core/Client/Client.php
```diff
@@ -766,7 +767,13 @@ public function removePlugin($plugin): ClientInterface
     public function createRequest(QueryInterface $query): Request
     {
         $event = new PreCreateRequestEvent($query);
-        $this->eventDispatcher->dispatch(Events::PRE_CREATE_REQUEST, $event);
+        if (Kernel::VERSION_ID >= 40300) {
+            // Support for symfony listeners which are using the old event name.
+            $this->eventDispatcher->dispatch($event, Events::PRE_CREATE_REQUEST);
+            $this->eventDispatcher->dispatch($event);
```
wickedOne: this fires the event twice without any form of sanity check so will break event dispatchers with one argument

composer.json
```diff
@@ -13,6 +13,7 @@
     ],
     "require": {
         "php": "^7.1",
+        "symfony/http-kernel": "^3.1 || ^4.0",
```
wickedOne: do we really need a dependency on the symfony kernel?
keep in mind that this is a php library rather than a symfony bundle.

mkalkbrenner: I second @wickedOne here!

mkalkbrenner:
This one is tricky :-(

I don't like the ` if (Kernel::VERSION_ID >= 40300) ` switches. What about following the recommendation of https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching instead and use the LegacyEventDispatcherProxy?

DemigodCode:
Didn't know the LegacyEventDispatcherProxy.
But now there's a new problem. This Proxy will break BC to symfony < 4.3.

mkalkbrenner:
> This Proxy will break BC to symfony < 4.3.

In which way?

DemigodCode:
Ah, my bad. have to check if LegacyEventDispatcherProxy is defined.

mkalkbrenner:
> have to check if LegacyEventDispatcherProxy is defined.

Exactly what I had in mind, but you were faster :-)

DemigodCode:
Okay, the legacy proxy provides BC from 5.0 to 4.3.
The proxy is in lower versions not available and therefor the different function signature problem between 4.2 and 4.3 still exists.

My biggest issue is that our monitoring is logging each deprecation and should do so. There are 4 deprecation warnings each site call. It's really messing things up here.

Any ideas?

DemigodCode:
Finally, there's one solution that works for all. I saw a similar one in other bundles.
The adventage here is, we should be sf 5.0 safe regarding events.

mkalkbrenner:
I'm not sure that this works in all cases. For example Drupal still uses symfony 3.

mkalkbrenner:
The event dispatcher doesn't depend on a symfony version. What about just doing this to ensure that the Proxy exists:
```
diff --git a/composer.json b/composer.json
index 8900aef3..67407857 100644
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "symfony/event-dispatcher": "^3.1 || ^4.0",
+        "symfony/event-dispatcher": "^4.3",
         "symfony/cache": "^3.1 || ^4.0",
         "ext-json": "*"
     },
```

DemigodCode:
Symfony 3 should still work. The code is checking for the proxy right now and uses the old way, if it doesn't exist.

mkalkbrenner:
Yes, I trust you. But I want to avoid all these checks. If my suggestion works, this would be mor elegant way and the minimum required symfony version will become 3.4.

DemigodCode:
Don't know if we are talking about different things, but travis says sf 3.3 is running

DemigodCode:
Ah, you mean the minimum required version will become 4.3? That's okay for me, too. Thought you don't want to break the BC to older versions.

mkalkbrenner:
https://github.com/symfony/event-dispatcher/blob/4.3/composer.json

symfony/event-dispatcher 4.3 requires at least symfony/dependency-injection 3.4
symfony/event-dispatcher 4.4 requires at least symfony/dependency-injection 4.4

So my patch from above will drop the support for any symfony version before 3.4.
But from my point of view that would be OK for solarium 5.1.0

mkalkbrenner:
Can you give it a try?
Otherwise I have to create a new PR :-(

.travis.yml
```diff
@@ -8,12 +8,11 @@ php:
   - nightly
 
 env:
-  - SYMFONY_VERSION=4.2.* SOLR_VERSION=8.1.1 SOLR_CLOUD=true
-  - SYMFONY_VERSION=4.2.* SOLR_VERSION=8.1.1 SOLR_CLOUD=false
-  - SYMFONY_VERSION=4.2.* SOLR_VERSION=7.7.2 SOLR_CLOUD=true
-  - SYMFONY_VERSION=4.2.* SOLR_VERSION=7.7.2 SOLR_CLOUD=false
-  - SYMFONY_VERSION=4.2.* SOLR_VERSION=6.6.6 SOLR_CLOUD=false
-  - SYMFONY_VERSION=3.3.* SOLR_VERSION=6.6.6 SOLR_CLOUD=false
```
mkalkbrenner: We should keep one test run against `SYMFONY_VERSION=3.4*` and add one against `SYMFONY_VERSION=4.4*`

composer.json
```diff
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "symfony/event-dispatcher": "^3.1 || ^4.0",
+        "symfony/event-dispatcher": "^4.0",
```
mkalkbrenner: it should be "symfony/event-dispatcher": "^4.3",

DemigodCode: sry, I'm getting a little bit careless :P

PR #685, opened 2019-07-23
Return the full body of the error

Firesphere:
The message is truncated, leaving out important details of what the actual error is.

Getting the body's content gives more clarity on what is actually going wrong.

### DIFF
--- a/src/Core/Client/Adapter/Guzzle.php
+++ b/src/Core/Client/Adapter/Guzzle.php
@@ -76,7 +76,7 @@ class Guzzle extends Configurable implements AdapterInterface
 
             return new Response((string) $guzzleResponse->getBody(), $responseHeaders);
         } catch (GuzzleException $e) {
-            $error = $e->getMessage();
+            $error = $e->getResponse()->getBody()->getContents();
             throw new HttpException("HTTP request failed, {$error}");
         }
     }


Firesphere:
I'll look at the failure later this week

Firesphere:
The problem with getMessage() is that it's truncated by Guzzle, which is what I am trying to resolve, because "Solr gave an error 400 because (truncated...)" is not useful to someone who doesn't have access to the Solr instance itself.

peter279k:
Hi @Firesphere, thanks for your reply.

To resolve this issue, I recommend that we can consider dispatching [RequestException](https://github.com/guzzle/guzzle/blob/master/src/Exception/RequestException.php).

For example, the original `Solr gave an error 400 because (truncated...)` message is about the request exception.

We can dispatch different Guzzle exceptions for different status codes.

mkalkbrenner:
@peter279k Can you provide a prototype?

src/Core/Client/Adapter/Guzzle.php
```diff
@@ -76,7 +76,7 @@ public function execute(Request $request, Endpoint $endpoint): Response
 
             return new Response((string) $guzzleResponse->getBody(), $responseHeaders);
         } catch (GuzzleException $e) {
-            $error = $e->getMessage();
+            $error = $e->getResponse()->getBody()->getContents();
```
peter279k: Firstly, we need to know about the Guzzle exceptions. These exceptions are as follows:

- GuzzleHttp\Exception\ClientException for 400-level errors
- GuzzleHttp\Exception\ServerException for 500-level errors
- GuzzleHttp\Exception\BadResponseException for both (it's their superclass)

The more detailed exception references are available [here](http://docs.guzzlephp.org/en/latest/quickstart.html#exceptions).

And we can change into following codes:

```php
......
use GuzzleHttp\Exception\ClientException;
use GuzzleHttp\Exception\ServerException;
......
try {
......
} catch (ClientException $e) {
......
    $error = $e->getResponse()->getBody()->getContents();
......
}
catch (ServerException $e) {
......
     $error = $e->getResponse()->getBody()->getContents();
}
......
```

This is for `Guzzle 6` version, but I notice that this package still support the `Guzzle 3.x`.

I think `Guzzle 3.x` version definition in `composer.json` can be removed because the official teams stop supporting this old version now.

mkalkbrenner: There's a dedicated Guzzle3 Adapter, so we don't need to care about it.

mkalkbrenner: What about this:
```
diff --git a/src/Core/Client/Adapter/Guzzle.php b/src/Core/Client/Adapter/Guzzle.php
index ef1d6485..0cccc70f 100644
--- a/src/Core/Client/Adapter/Guzzle.php
+++ b/src/Core/Client/Adapter/Guzzle.php
@@ -3,6 +3,7 @@
 namespace Solarium\Core\Client\Adapter;
 
 use GuzzleHttp\Client as GuzzleClient;
+use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Exception\GuzzleException;
 use GuzzleHttp\RequestOptions;
 use Solarium\Core\Client\Endpoint;
@@ -76,7 +77,15 @@ class Guzzle extends Configurable implements AdapterInterface
 
             return new Response((string) $guzzleResponse->getBody(), $responseHeaders);
         } catch (GuzzleException $e) {
+            // Guzzle truncates the error message, therefore we try to get the full message from Solr below
             $error = $e->getMessage();
+            if ($e instanceof BadResponseException) {
+                try {
+                    $error = $e->getResponse()->getBody()->getContents();
+                } catch (\Exception $e) {
+                    // Fall back to the short message
+                }
+            }
             throw new HttpException("HTTP request failed, {$error}");
         }
     }

```

peter279k: @mkalkbrenner, thanks for your reply.

It looks fine to use the  `BadResponseException` super class to catch the `ServerException` and `ClientException`.

After looking at the `Guzzle 3.x` exception [references](https://github.com/guzzle/guzzle/tree/v3.8.1/src/Guzzle/Http/Exception), it's very diffrernt from `Guzzle 6.x`.

The old `Guzzle 3.x` has the different namespace for the exception.

mkalkbrenner: As already mentioned, we shouldn't care about Guzzle 3 anymore, just leave it as it is.

I'll open an new PR for my suggestion.

peter279k: Ok. That's fine.

Firesphere:
You folks are awesome in solving my issue! You did a lot better than my initial "just get'r done" fix (admitted, I simply hardly had any time to properly have a look).

Thanks for understanding my issue and moving so fast!

PR #704, opened 2019-09-20
Update Pivot.php

szabizs:
Provide essential pivot methods: setLimit() and getLimit()


mkalkbrenner:
Thanks, but we won't touch 3.x anymore.

PR #709, opened 2019-10-04
fix: MoreLikeThis handler must be "select"

andiechang:
Solr version 8.1.1
MoreLikeThis handler must be "select"

### DIFF
--- a/src/QueryType/MoreLikeThis/Query.php
+++ b/src/QueryType/MoreLikeThis/Query.php
@@ -23,7 +23,7 @@ class Query extends SelectQuery
      * @var array
      */
     protected $options = [
-        'handler' => 'mlt',
+        'handler' => 'select',
         'resultclass' => Result::class,
         'documentclass' => Document::class,
         'query' => '*:*',


mkalkbrenner:
The handler could be named freely in solrconf.xml. "mlt" is a common name.
But "select" sounds wrong. "select" is the suggested default request handler for searches. It might contain a MoreLikeThis **component** but should not be a MoreLikeThis _request handler_.

> The second is to use it as a search component. This is less desirable since it performs the MoreLikeThis analysis on every document returned. This may slow search results.

https://lucene.apache.org/solr/guide/8_1/morelikethis.html

andiechang:
> The handler could be named freely in solrconf.xml. "mlt" is a common name.
> But "select" sounds wrong. "select" is the suggested default request handler for searches. It might contain a MoreLikeThis **component** but should not be a MoreLikeThis _request handler_.
> 
> > The second is to use it as a search component. This is less desirable since it performs the MoreLikeThis analysis on every document returned. This may slow search results.
> 
> https://lucene.apache.org/solr/guide/8_1/morelikethis.html

The handler "select" here is use to request Solr API like: "http://localhost:8983/solr/techproducts/select?q=id:SP2514N&wt=xml".

If set handler to "mlt", the request will change to like this: "http://localhost:8983/solr/techproducts/mlt?q=id:SP2514N&wt=xml". 
And this request will get an error: 
```
Solr HTTP error: HTTP request failed,
HTTP ERROR 404
Problem accessing /solr/techproducts/mlt.
Reason: Not Found
```

In official Doc: https://solarium.readthedocs.io/en/stable/queries/morelikethis-query/
Default value of "handler" is "select".

mkalkbrenner:
Techproducts is an example that doesn't have a MoreLikeThisHandler!

The Techproducts example contains a MoreLikeThisComonent:
`<searchComponent name="mlt" class="solr.MoreLikeThisComponent" />`

To talk to this component you should use a Select Query:
https://solarium.readthedocs.io/en/stable/queries/select-query/building-a-select-query/components/morelikethis-component/

If a MoreLikeThisHandler is present, it's best practice to name it `/mlt`.

But you're right about the documentation. The handler in the documentation is wrong!

andiechang:
I see!!!
There are two way to use MoreLikeThis.
"[Building a MLT query](https://solarium.readthedocs.io/en/stable/queries/morelikethis-query/)"
"[Building a select query to talk to MoreLikeThis component](https://solarium.readthedocs.io/en/stable/queries/select-query/building-a-select-query/components/morelikethis-component/)"

@mkalkbrenner  Thanks!!!

PR #718, opened 2019-11-23
Added Symfony 5 support

jzawadzki:
#SymfonyHackday


mkalkbrenner:
Thanks, but fixed differently by ef308267 because of #723

PR #722, opened 2019-12-05
PHP 7.4 deprecation: implode() $glue must be first paramter

### LINKED ISSUE #720: Deprecation warning in PHP 7.4
After upgrading to PHP 7.4, we are receiving the following deprecated warning.

> implode(): Passing glue string after array is deprecated. Swap the parameters in ../vendor/solarium/solarium/library/Solarium/Core/Query/AbstractRequestBuilder.php

This warning causes a ReactJS page to fail completely, as it requires the response to be clean.

ejunker:
Passing the $glue and $pieces parameters in reverse order to implode has been deprecated since PHP 7.4; $glue should be the first parameter and $pieces the second

fixes #720

### DIFF
--- a/src/Core/Query/AbstractRequestBuilder.php
+++ b/src/Core/Query/AbstractRequestBuilder.php
@@ -68,7 +68,7 @@ abstract class AbstractRequestBuilder implements RequestBuilderInterface
             }
 
             if (is_array($paramValue)) {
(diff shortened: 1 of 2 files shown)


mkalkbrenner:
Thanks! But this issue was already fixed by #711

PR #654, opened 2019-02-23
range pivot

wickedOne:
hi thre,

i needed to implement a [range pivot](https://lucene.apache.org/solr/guide/6_6/faceting.html#Faceting-CombiningFacetQueriesAndFacetRangesWithPivotFacets) and could not find a way to implement this using solarium so i made some changes to be able to.

as you can see this pr is targeting the 3.x branch as that's the one we're currently usinng.

i'm willing to target the 4.x branch and write some unittests for this feature, but
- are you interested in this feature at all?
- will it be available in the 3.x branch as i donn't know when we've got time to upgrade

the givenn example can now be achieved by:
```php
$facetSet = $query->getFacetSet();
$facet = $facetSet->createFacetRange(['key' => 'manufacturedate_dt', 'tag' => 'r1']);

$facet->setField('manufacturedate_dt');
$facet->setStart('2006-01-01T00:00:00Z');
$facet->setEnd('NOW/YEAR'));
$facet->setGap('+1YEAR');
$facet->setPivot(['cat', 'inStock']);
```

PR #726, opened 2019-12-08
range pivot

wickedOne:
this one replaces https://github.com/solariumphp/solarium/pull/654

these addition enables one to add a pivot to their range query as described over here: https://lucene.apache.org/solr/guide/6_6/faceting.html#Faceting-CombiningFacetQueriesAndFacetRangesWithPivotFacets.

the given example can now be achieved by:
```php
$facetSet = $query->getFacetSet();
$facet = $facetSet->createFacetRange(['key' => 'manufacturedate_dt', 'tag' => 'r1']);

$facet->setField('manufacturedate_dt');
$facet->setStart('2006-01-01T00:00:00Z');
$facet->setEnd('NOW/YEAR'));
$facet->setGap('+1YEAR');
$facet->setPivot(['cat', 'inStock']);
```

### DIFF
--- a/src/Component/Facet/Range.php
+++ b/src/Component/Facet/Range.php
@@ -46,6 +46,24 @@ class Range extends AbstractRange
     }
 
     /**
+     * @param string $tag
+     *
+     * @return \Solarium\Core\Configurable
+     */
+    public function setTag($tag)
+    {
+        return $this->setOption('tag', $tag);
+    }
+
+    /**
+     * @return string
+     */
+    public function getTag()
+    {
+        return $this->getOption('tag');
+    }
+
+    /**
      * Initialize options.
      *
      * Several options need some extra checks or setup work, for these options
--- a/src/Component/Facet/Range.php
+++ b/src/Component/Facet/Range.php
@@ -46,24 +46,6 @@ class Range extends AbstractRange
     }
 
     /**
-     * @param string $tag
-     *
-     * @return \Solarium\Core\Configurable
-     */
-    public function setTag($tag)
-    {
-        return $this->setOption('tag', $tag);
-    }
-
-    /**
-     * @return string
-     */
-    public function getTag()
-    {
-        return $this->getOption('tag');
-    }
-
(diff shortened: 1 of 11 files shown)


src/Component/Facet/Range.php
```diff
@@ -48,6 +48,42 @@ public function getMinCount(): ?int
...
+    {
+        return $this->getOption('tag');
+    }
+
+    /**
+     * @param string|array $pivot
+     *
+     * @return \Solarium\Core\Configurable
+     */
+    public function setPivot($pivot)
```
mkalkbrenner: Don't these additions apply to JsonRange as well and should therefore be moved to the AbstractRange base class?

src/Component/Facet/Range.php
```diff
@@ -48,6 +48,42 @@ public function getMinCount(): ?int
         return $this->getOption('mincount');
     }
 
+    /**
+     * @param string $tag
+     *
+     * @return \Solarium\Core\Configurable
+     */
+    public function setTag($tag)
```
mkalkbrenner: According to https://lucene.apache.org/solr/guide/6_6/faceting.html tags are generic facet parameters like exludes.
Therefore we should not implement them specifically for range facets. We should follow or improve or extend the pattern of The ExludeTagsInterface and ExcludeTagsTrait combo.

wickedOne: doesn't that imply solarium needs a complete rework of how it deals with local params?

src/Component/Facet/Range.php
wickedOne: moved to the abstract range...

src/Component/Facet/Range.php
wickedOne: https://github.com/solariumphp/solarium/issues/735

wickedOne: can be modified once https://github.com/solariumphp/solarium/pull/736 is reviewed and merged

PR #733, opened 2020-01-02
select query builder

wickedOne:
not sure whether this would be a useful addition to this library so i'm just going to leave this here for you guys to decide.

the intention is to ease the writing of solr filter queries by using a builder which, as you can see, is heavily inspired by doctrine's criteria object.

this pr has no integration with the select query type for now, but should be fairly simple to create a method which takes a filter builder instance as arguments and iterates over the expressions and add them as filter queries.

see the phpunit test for various examples, writing filter queries will be as easy as
```php
$filter = FilterBuilder::create()
    ->where(FilterBuilder::expr()->eq('title', 'foo'))
    ->andWhere(FilterBuilder::expr()->like('description, 'bar*'))
;
```

### DIFF
--- a/src/Builder/Select/QueryExpressionVisitor.php
+++ b/src/Builder/Select/QueryExpressionVisitor.php
@@ -1,0 +1,190 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Solarium\Builder\Select;
+
+use Solarium\Builder\AbstractExpressionVisitor;
+use Solarium\Builder\Comparison;
+use Solarium\Builder\CompositeComparison;
+use Solarium\Builder\ExpressionInterface;
+use Solarium\Builder\Value;
+use Solarium\Core\Query\Helper;
+use Solarium\Exception\RuntimeException;
+
+/**
+ * Query Expression Visitor.
+ *
+ * @author wicliff <wicliff.wolda@gmail.com>
+ */
+class QueryExpressionVisitor extends AbstractExpressionVisitor
+{
+    /**
+     * @var Helper
+     */
+    private $helper;
+
+    /**
+     * Constructor.
+     */
+    public function __construct()
+    {
+        $this->helper = new Helper();
+    }
+
+    /**
+     * @param \Solarium\Builder\ExpressionInterface $expression
+     *
+     * @throws \Solarium\Exception\RuntimeException
+     *
+     * @return mixed|string
+     */
+    public function walkExpression(ExpressionInterface $expression)
+    {
+        $field = $expression->getField();
+        $value = $expression->getValue()->getValue();
+
+        switch ($expression->getOperator()) {
+            case Comparison::EQ:
+            case Comparison::NEQ:
+                $strValue = $this->valueToString($value, ',', '"');
+
+                if ($value instanceof \DateTime) {
+                    $strValue = sprintf('[%1$s TO %1$s]', $strValue);
+                }
+
+                $not = (Comparison::NEQ === $expression->getOperator()) ? '-' : '';
+
+                return sprintf('%s%s:%s', $not, $field, $strValue);
+            case Comparison::GT:
+                return sprintf('%s:{%s TO *]', $field, $this->valueToString($value));
+            case Comparison::GTE:
+                return sprintf('%s:[%s TO *]', $field, $this->valueToString($value));
+            case Comparison::LT:
+                return sprintf('%s:[* TO %s}', $field, $this->valueToString($value));
+            case Comparison::LTE:
+                return sprintf('%s:[* TO %s]', $field, $this->valueToString($value));
+            case Comparison::RANGE:
+                if (\is_array($value)) {
+                    if (2 === \count($value)) {
+                        return sprintf('%s:[%s TO %s]', $field, $this->valueToString($value[0]), $this->valueToString($value[1]));
+                    }
+
+                    if (1 === \count($value)) {
+                        return sprintf('%s:[%s TO *]', $field, $this->valueToString($value[0]));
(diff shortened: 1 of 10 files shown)


src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @return string
+     */
+    private function typedValueToString($value, string $quote = '', $escape = true): string
+    {
+        if (null === $value) {
+            return '[* TO *]';
+        }
+
+        if ($value instanceof \DateTime) {
```
mkalkbrenner: Have a look at Helper::formatDate and leverage it

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,178 @@
...
+use Solarium\Exception\RuntimeException;
+
+/**
+ * Query Expression Visitor.
+ *
+ * @author wicliff <wicliff.wolda@gmail.com>
+ */
+class QueryExpressionVisitor extends AbstractExpressionVisitor
+{
+    private const PATTERN = '/( |\+|-|&&|\|\||!|\(|\)|\{|}|\[|]|\^|"|~|\*|\?|:|\/|\\\)/';
```
mkalkbrenner: leverage Helper::escapeTerm()

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,178 @@
...
+        if (null === $value) {
+            return '[* TO *]';
+        }
+
+        if ($value instanceof \DateTime) {
+            return sprintf('%s%s', strstr($value->format(\DateTime::ATOM), '+', true), 'Z');
+        }
+
+        if (true === $escape && \is_string($value)) {
+            $value = preg_replace(self::PATTERN, '\\\$1', $value);
```
mkalkbrenner: leverage Helper::escapeTerm()

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,178 @@
...
+     * @param mixed  $value
+     * @param string $quote
+     * @param bool   $escape
+     *
+     * @return string
+     */
+    private function typedValueToString($value, string $quote = '', $escape = true): string
+    {
+        if (null === $value) {
+            return '[* TO *]';
```
mkalkbrenner: This will fail on location data. Use Helper::rangeQuery().

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,178 @@
...
+                $strValue = $this->valueToString($value, ',', '"');
+
+                if ($value instanceof \DateTime) {
+                    $strValue = sprintf('[%1$s TO %1$s]', $strValue);
+                }
+
+                $not = (Comparison::NEQ === $expression->getOperator()) ? '-' : '';
+
+                return sprintf('%s%s:%s', $not, $field, $strValue);
+            case Comparison::GT:
```
mkalkbrenner: In drupal's Search API Solr module we have a very similar logic compared to all these "cases".
We should compare both implementations and isolate it here / move it to Helper in order to leverage it from Search API Solr.

Have a look at https://git.drupalcode.org/project/search_api_solr/blob/8.x-3.x/src/Plugin/search_api/backend/SearchApiSolrBackend.php#L2925

mkalkbrenner: I think we might have solved some more edge cases there. In general it would be a good choice to move that stuff to solarium and to leverage it form Search API Solr, regardless if as methods or the new Query Builder.

src/Builder/Select/QueryExpressionVisitor.php
wickedOne: looks like the helper returns the same for an empty value?
https://github.com/solariumphp/solarium/blob/d6d1adb67a2cc905a3783bf57cec152d495e6fcc/src/Core/Query/Helper.php#L183-L196

wickedOne:
converted this one to draft as i'm somewhat struggling with the concept of adding the possibility to add filter query options in the process.

i would understand the need for this, but am wondering whether it would make this too complex and we should rebrand this as a simple query builder (i.e. wiithout support for, for instance, setting local parameters for the filter queries).

any input is appreciated

mkalkbrenner:
> converted this one to draft as i'm somewhat struggling with the concept of adding the possibility to add filter query options in the process.

Filter queries are an essential concept in Solr. You should use them as much as possible instead of creating long queries with boolean operators because filter queries get cached individually in Solr.

wickedOne:
> Filter queries are an essential concept in Solr. You should use them as much as possible instead of creating long queries with boolean operators because filter queries get cached individually in Solr.

yes, the intention of this builder to ease the creation of multiple filter queries in one go. the only thing which, with the current state of this pr, isn't possible is to add options / parameters to the query.

for instance
```
fq=quantity_in_stock:[5 TO *]
``` 
could be accomplished by 

```php
FilterBuilder::create()
    ->where(FilterBuilder::expr()->range('quantity_in_stock', [5, *]))
;
```
however, if you want your filter query to end up like
```
fq={!cache=false}quantity_in_stock:[5 TO *]
```
there's currently no way to do that with this query builder.

so the question is: do we keep it simple and direct users to the current way of constructing filter queries or do we want to put effort in being able to add options / parameters to filter queries?

i personally am ok with keeping it simple, so what do you think?

mkalkbrenner:
Keep it simple :-)

PR #736, opened 2020-01-04
local parameters

### LINKED ISSUE #735: rework local params
we might want to move all local params logic to it's own trait so it can be easily recycled across facets etc.

https://lucene.apache.org/solr/guide/8_3/local-parameters-in-queries.html

wickedOne:
a re-implementation to add local parameters to solr queries

fixes #735


src/Core/Query/LocalParameters/LocalParameter.php
```diff
@@ -0,0 +1,152 @@
...
+
+    public const TYPE_MAX = 'max';
+
+    public const TYPE_MEAN = 'mean';
+
+    public const TYPE_DEFAULT_FIELD = 'df';
+
+    public const TYPE_QUERY_FIELD = 'qf';
+
+    public const TYPE = 'type';
```
mkalkbrenner: Wouldn't TYPE_TYPE be more consistent?

src/Core/Query/LocalParameters/LocalParametersTrait.php
```diff
@@ -0,0 +1,222 @@
...
+     * Init local parameters.
+     */
+    protected function initLocalParameters(): void
+    {
+        $this->localParameters = new LocalParameters();
+
+        foreach ($this->options as $name => $value) {
+            switch ($name) {
+                case 'exclude':
+                case LocalParameter::PARAMETER_MAP[LocalParameter::TYPE_EXCLUDE]:
```
mkalkbrenner: this list is rather long. What about collapsing the 'cases'?
```php
$this->getLocalParameters()->addValues($name, $value);
```

src/Core/Query/LocalParameters/LocalParameter.php
wickedOne: yes you're absoletely right; will modify.

src/Core/Query/LocalParameters/LocalParametersTrait.php
wickedOne: i see your point, but that would mean the ``setValue($name, $value)``, ``addValues($name, $value)``, etc. methods have to be made public.

in my opinion this would "encourage" people to add values to the local parameters directly without solarium being able to enforce the actual method called or to (for now) have legacy option names set the correct local parameter or to have a different scenario for any given value (e.g. do the ``is_array()`` check and explode if necessary).

nevertheless, in order to simplify the switch i'd like to suggest to seperate the legacy parameter names ('exclude', 'tag', 'stats') and have them call the appropriate method until these options are deprecated, group the options which need an array check and group those remaining. make the addValues and addValues methods public and have them throw an exception when a parameter name is passed which is not in the ``LocalParameter::PARAMETER_MAP``. would that work for you?

Issue #550, opened 2018-01-19
Use php-http instead of custom client implementations


thePanz:

thomascorthals:
This makes #397 a moot point.

dmaicher:
I guess with the added `Psr18Adapter` this can be closed now?

PR #738, opened 2020-01-17
zend framework is now called laminas

wickedOne:
replaced the zendframework/zend-http dev requirement with
laminas/laminas-http as zendframework is abandoned

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,10 +19,10 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.16",
         "guzzlehttp/guzzle": "^3.8 || ^6.2",
+        "laminas/laminas-http": "^2.8",
         "php-coveralls/php-coveralls": "^2.1",
         "phpunit/phpunit": "^8.0",
-        "squizlabs/php_codesniffer": "^3.4",
-        "zendframework/zend-http": "^2.8"
+        "squizlabs/php_codesniffer": "^3.4"
     },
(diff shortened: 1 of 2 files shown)


wickedOne:
fixed the style violation @ https://github.com/solariumphp/solarium/pull/738/commits/dfc42631b607bf949c69cfd820cb39652d1be6c6

after moving to 5.2 we might want to consider user deprecation for this namespace and rename it ``LaminasHttp`` or do you prefer dropping all custom adapters and implement #550 instead?

mkalkbrenner:
> after moving to 5.2 we might want to consider user deprecation for this namespace and rename it LaminasHttp or do you prefer dropping all custom adapters and implement #550 instead?

What's your opinion about it? I hate to have multiple adapters. From my point of view, one would be enough. Isn't guzzle an abstraction already? but #550 might be OK as well.

wickedOne:
back when those adapters were created it was great to support multiple clients, now with PSR-18 around i think #550 is the way to go. 

this way solarium no longer dictates the http client to be used in projects using solarium while we no longer have to maintain dependencies like in this pr: win win :-)

more than willing to give #550 a go in the near future which should be able to live amongst the current adapters which then can be deprecated in v6.x

what do you think?

mkalkbrenner:
I agree!

dmaicher:
This can be closed now that we have the `Psr18Adapter`, or? :blush:

PR #740, opened 2020-01-21
Added support for useConfiguredElevatedOrder to Query Elevation

thomascorthals:
Added support for useConfiguredElevatedOrder in Query Elevation Component, including test coverage and documentation.

### DIFF
--- a/src/Component/RequestBuilder/QueryElevation.php
+++ b/src/Component/RequestBuilder/QueryElevation.php
@@ -32,6 +32,7 @@ class QueryElevation implements ComponentRequestBuilderInterface
         $request->addParam('enableElevation', $component->getEnableElevation());
         $request->addParam('forceElevation', $component->getForceElevation());
         $request->addParam('exclusive', $component->getExclusive());
+        $request->addParam('useConfiguredElevatedOrder', $component->getUseConfiguredElevatedOrder());
         $request->addParam('markExcludes', $component->getMarkExcludes());
 
         // add overrides for pre-configured elevations
--- a/docs/queries/select-query/building-a-select-query/components/query-elevation-component.md
+++ b/docs/queries/select-query/building-a-select-query/components/query-elevation-component.md
@@ -3,15 +3,16 @@ Query Elevation is a Solr component that lets you configure the top results for 
 Options
 -------
 
-| Name            | Type    | Default value | Description                                                                                                                                            |
-|-----------------|---------|---------------|--------------------------------------------------------------------------------------------------------------------------------------------------------|
-| transformers    | string  | [elevated]    | Comma separated list of transformers to annotate each document. The [elevated] transformer tells whether or not the document was elevated.             |
-| enableElevation | boolean | null          | For debugging it may be useful to see results with and without elevation applied. To get results without elevation, use false.                         |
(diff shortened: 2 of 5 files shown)


src/Component/RequestBuilder/QueryElevation.php
```diff
@@ -32,6 +32,7 @@ public function buildComponent(ConfigurableInterface $component, Request $reques
         $request->addParam('enableElevation', $component->getEnableElevation());
         $request->addParam('forceElevation', $component->getForceElevation());
         $request->addParam('exclusive', $component->getExclusive());
+        $request->addParam('useConfiguredElevatedOrder', $component->getUseConfiguredElevatedOrder());
```
wickedOne: thanks for your contribution!

not sure whether the documentation is vague or this part might be wrong; the documentation states:
> You can force set useConfiguredElevatedOrder during runtime by supplying it as a request parameter

https://lucene.apache.org/solr/guide/8_3/the-query-elevation-component.html#the-useconfiguredelevatedorder-parameter

doesn't that mean this parameter should only be added when ``$component->getUseConfiguredElevatedOrder()``  is true?

thomascorthals: The documentation is vague. I had to run some tests to make sense of it. The parameter only applies if `forceElevation=true`.

`&elevateIds=3,1&forceElevation=true&sort=id%20ASC`
```json
{"id":3, "[elevated]":true}, {"id":1,  "[elevated]":true}, {"id":2, "[elevated]":false}, {"id":4, "[elevated]":false}, {"id":5, "[elevated]":false}
```

`&elevateIds=3,1&forceElevation=true&useConfiguredElevatedOrder=true&sort=id%20ASC`
```json
{"id":3, "[elevated]":true}, {"id":1,  "[elevated]":true}, {"id":2, "[elevated]":false}, {"id":4, "[elevated]":false}, {"id":5, "[elevated]":false}
```

`&elevateIds=3,1&forceElevation=true&useConfiguredElevatedOrder=false&sort=id%20ASC`
```json
{"id":1, "[elevated]":true}, {"id":3,  "[elevated]":true}, {"id":2, "[elevated]":false}, {"id":4, "[elevated]":false}, {"id":5, "[elevated]":false}
```

Default behaviour without the parameter is the same as setting it explicitly to true. Setting it explicitly to false changes this behaviour. It must definitely be added when `$component->getUseConfiguredElevatedOrder()` is false and there's no harm in adding it when it's true. I would alway add it when explicitly set and omit when null. That's how parameters are handled in most cases.

[Solr's source code](https://github.com/apache/lucene-solr/blob/master/solr/solrj/src/java/org/apache/solr/common/params/QueryElevationParams.java) does a better job describing this behaviour than the documentation does. Although they fail to mention it also applies to `elevateIds`.

```java
  /**
   * When multiple docs are elevated, should their relative order be the order in the configuration file or should
   * they be subject to whatever the sort criteria is?  True by default.
   */
  String USE_CONFIGURED_ELEVATED_ORDER = "useConfiguredElevatedOrder";
```

wickedOne: thanks for your extensive explanation on this part!

PR #742, opened 2020-02-01
Fixed bug with special characters in Managed Resources

thomascorthals:
It's possible to add stopwords/synonyms with characters that are reserved per the URI specification ([RFC 3986](https://tools.ietf.org/html/rfc3986#section-2.2)). This creates a problem with the URI for the REST resource when checking their existence or trying to delete them. Percent-encoding the term once isn't sufficient, it takes a second pass to percent-encode the percent signs. (Even then, it's impossible to get it to work with a forward slash in the term. But that's a webserver thing.)

A similar but distinct problem arises with spaces. It happens to be solved by the double percent-encoding. Percent-encoding only once would have worked too in this case.

I've added tests to prevent regressions.


src/QueryType/ManagedResources/RequestBuilder/Stopwords.php
```diff
@@ -53,10 +53,11 @@ protected function buildCommand(Request $request, AbstractCommand $command): sel
                 $request->setRawData($command->getRawData());
                 break;
             case StopwordsQuery::COMMAND_DELETE:
-                $request->setHandler($request->getHandler().'/'.$command->getTerm());
+                // reserved characters in a REST resource name need to be encoded twice to make it through the servlet
+                $request->setHandler($request->getHandler().'/'.rawurlencode(rawurlencode($command->getTerm())));
```
mkalkbrenner: I agree that we missed to encode the term! But the double encoding surprises me. I can't believe that this is required.

tests/QueryType/ManagedResources/RequestBuilder/StopwordsTest.php
```diff
@@ -99,4 +99,30 @@ public function testExists()
         $this->assertEquals($term, $command->getTerm());
         $this->assertEquals('', $command->getRawData());
     }
+
+    public function testReservedCharacters()
```
mkalkbrenner: These are just unit tests that don't prove that the double encoding really works with Solr. We already have Integration tests that query a real Solr server on travis. I think such a test would be preferable here.

src/QueryType/ManagedResources/RequestBuilder/Stopwords.php
thomascorthals: I first ran into the problem with a SPACE character. That was fixed with encoding once. I also thought that would fix it for all special characters, but it empirically doesn't. It can be reproduced from the command line, so it's not PHP specific. You'll notice that both `C++` and `C%2B%2B` get interpreted as `C  ` (C SPACE SPACE). `C%23anything` acts like you'd expect from a fragment identifier, which it clearly isn't.

```
thomas@testbox:~$ curl -X PUT -H 'Content-type:application/json' --data-binary '{"class":"org.apache.solr.rest.schema.analysis.ManagedSynonymGraphFilterFactory$SynonymManager"}' "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test"
{
  "responseHeader":{
    "status":0,
    "QTime":0}}
thomas@testbox:~$ curl -X PUT -H 'Content-type:application/json' --data-binary '["C++","C plus plus"]' "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test"
{
  "responseHeader":{
    "status":0,
    "QTime":0}}
thomas@testbox:~$ curl -X PUT -H 'Content-type:application/json' --data-binary '["C#","C sharp"]' "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test"
{
  "responseHeader":{
    "status":0,
    "QTime":0}}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test"
{
  "responseHeader":{
    "status":0,
    "QTime":0},
  "synonymMappings":{
    "initArgs":{"ignoreCase":false},
    "initializedOn":"2020-02-01T11:48:35.235Z",
    "updatedSinceInit":"2020-02-01T11:49:01.280Z",
    "managedMap":{
      "C plus plus":["C plus plus",
        "C++"],
      "C sharp":["C sharp",
        "C#"],
      "C#":["C sharp",
        "C#"],
      "C++":["C plus plus",
        "C++"]}}}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test/C++"
{
  "responseHeader":{
    "status":404,
    "QTime":0},
  "error":{
    "metadata":[
    "msg":"C   not found in /schema/analysis/synonyms/test",
    "code":404}}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test/C%2B%2B"
{
  "responseHeader":{
    "status":404,
    "QTime":0},
  "error":{
    "metadata":[
    "msg":"C   not found in /schema/analysis/synonyms/test",
    "code":404}}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test/C%252B%252B"
{
  "responseHeader":{
    "status":0,
    "QTime":0},
  "C++":["C plus plus",
    "C++"]}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test/C%23"
{
  "responseHeader":{
    "status":404,
    "QTime":0},
  "error":{
    "metadata":[
    "msg":"C not found in /schema/analysis/synonyms/test",
    "code":404}}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test/C%23anything"
{
  "responseHeader":{
    "status":404,
    "QTime":0},
  "error":{
    "metadata":[
    "msg":"C not found in /schema/analysis/synonyms/test",
    "code":404}}
thomas@testbox:~$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/test/C%2523"
{
  "responseHeader":{
    "status":0,
    "QTime":0},
  "C#":["C sharp",
    "C#"]}
```

I couldn't find an 'official' explanation for this behaviour. Whenever a problem like this pops up on Stack Overflow ([example](https://stackoverflow.com/questions/14359305/mvc-web-api-routing-fails-when-url-contains-encoded-ampersand)), the advice is to use query parameters instead of a URI path. Not very useful if you're only developing on the client side.

The [/ problem](https://stackoverflow.com/questions/3235219/urlencoded-forward-slash-is-breaking-url) seems to be beyond hope.

tests/QueryType/ManagedResources/RequestBuilder/StopwordsTest.php
thomascorthals: A real Solr test would be preferable. I just felt some test was necessary. The double encoding (if we settle on it) is so non-obvious that it's a prime candidate for regressions.

src/QueryType/ManagedResources/RequestBuilder/Stopwords.php
thomascorthals: [SOLR-6853](https://issues.apache.org/jira/browse/SOLR-6853) is the bug report for this behaviour. It was created in 2014 but remains unresolved.

thomascorthals: > [SOLR-6853](https://issues.apache.org/jira/browse/SOLR-6853) is the bug report for this behaviour. It was created in 2014 but remains unresolved.

I've added my above example to the Solr issue as well.

tests/QueryType/ManagedResources/RequestBuilder/StopwordsTest.php
mkalkbrenner: @thomascorthals Can you solve the conflicts with master and an integration test?

tests/Integration/Fixtures/Techproducts/gb18030-example.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="GB18030"?>
```
mkalkbrenner, line 1: I assume, the PR isn't finished yet, right?
Posting the test data via solarium would be a great addition to the integration tests :-)

thomascorthals: My bad. I thought I had those changes stashed away. Wasn't meant to be added just yet.

I was working on importing them as suggested in #751 by yourself. Didn't get to finish it yet, work (and — it shows — sleep) schedule have been impacted by current world affairs.

I'll remove them for now, and commit them with the PR they were meant for.

mkalkbrenner:
I think we could add an integration test now. It is important that we notice if Solr fixes the issue and our workaround will cause trouble.

thomascorthals:
> I think we could add an integration test now. It is important that we notice if Solr fixes the issue and our workaround will cause trouble.

I've added resources with reserved characters to the integration tests and included an additional integration test that compares the request builder that works with Solr against an RFC 3986 compliant request builder that triggers SOLR-6853. When Solr fixes that behaviour, this test will fail even if they still support the workaround for BC. I've made sure to mention SOLR-6853 in comments throughout the source code to make it easy to track down the work we did and the tests we wrote in case we ever have to remove the workaround. As with all Solr issues I've encountered with Solarium, I'm watching it on their JIRA too.

Didn't think of testing reserved characters in the resource name before. It turns out the same double percent-encoding is necessary (no surprise there as I suspect the servlet is actually to blame for that). However, there is one difference that did surprise me: a resource name _can_ include a properly (double) percent-encoded slash while a term _can't_. I've reported this behaviour with the `curl` commands to reproduce in [SOLR-6853](https://issues.apache.org/jira/browse/SOLR-6853). (@dkd-kaehm Since you've ported the fix for terms, this might interest you as well.)

mkalkbrenner (approved):

Issue #744, opened 2020-02-01
Allow disabling cache for FilterQuery?


dmaicher:
I have a use case where a filter query consists of a union of multiple sub-filters:

Like

`fq=(filter(field:value1) OR filter(field:value2) OR filter(field:value3))`

This works fine with the current `FilterQuery` class in Solarium.

But now I would like to disable caching for this top-level filter (as caching the sub-filters is enough in my case).

Like

`fq={!cache=false}(filter(field:value1) OR filter(field:value2) OR filter(field:value3))`

I don't think this is currently doable, right?

dmaicher:
It seems on master its somewhat doable since https://github.com/solariumphp/solarium/pull/736

```php
$filterQuery->getLocalParameters()->offsetSet('cache', new class implements LocalParameterInterface {
    public function __toString(): string
    {
        return 'cache=false';
    }

    public function getType(): string
    {
        return 'cache';
    }

    public function getValues(): array
    {
        return [];
    }

    public function clearValues(): LocalParameterInterface
    {
        return $this;
    }

    public function addValue($value): LocalParameterInterface
    {
        return $this;
    }

    public function addValues(array $values): LocalParameterInterface
    {
        return $this;
    }

    public function removeValue($value): LocalParameterInterface
    {
        return $this;
    }
});
```

But this seems overly complicated and hacky :yum: 

For me this interface `LocalParameterInterface` seems to be not a good fit in this case? Only one method (`__toString`) is actually needed and used.

ping @wickedOne

mkalkbrenner:
I think we only need to add the `cache` type to the LocalParameter class and corresponding getter and setter to the FilterQuery calss. Then it would be
```php
$filterQuery->setCache('false');
```

wickedOne:
yes that's rather hacky and most of all (at least from what i could find) the ``cache`` and ``costs`` parameters are not local parameters but are parameters explicitly for filter queries.

though i haven't had the time to test this, but i'm not sure what will happen if you end up with...
```
facet.range={!cache=false cost=50}
```
which you will enable by making it a local parameter.

nevertheless they need to be merged in the same section as the local parameters in order for it to work. so we might want to come up with something more elegant which extends the ``LocalParameters`` class and overrides the ``render`` method for example?

PR #746, opened 2020-02-04
allow disabling cache for filter queries

dmaicher:
Fixes https://github.com/solariumphp/solarium/issues/744


mkalkbrenner:
@dmaicher Please excuse. It seems that we worked on the same issue at the same time :-(
I think my PR is more kind of feature complete. Could you please review and test it?
see #748

dmaicher:
@mkalkbrenner no problem. Thanks for working on it :+1:

PR #748, opened 2020-02-04
fixes #744, deprecate erroneous Helper::cacheControl()

### LINKED ISSUE #744: Allow disabling cache for FilterQuery?

mkalkbrenner:
add cache and cost local params

### DIFF
--- a/src/Core/Query/LocalParameters/LocalParameters.php
+++ b/src/Core/Query/LocalParameters/LocalParameters.php
@@ -745,6 +745,70 @@ class LocalParameters implements \ArrayAccess
     }
 
     /**
+     * @param bool $cache
+     *
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return $this
+     */
+    public function setCache(bool $cache): self
+    {
+        return $this->addValue(LocalParameter::TYPE_CACHE, $cache ? 'true' : 'false');
+    }
+
+    /**
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return $this
+     */
+    public function clearCache(): self
+    {
+        return $this->clearValues(LocalParameter::TYPE_CACHE);
+    }
+
+    /**
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return array
+     */
+    public function getCache(): array
+    {
+        return $this->getValues(LocalParameter::TYPE_CACHE);
+    }
+
+    /**
+     * @param int $cost
+     *
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return $this
+     */
+    public function setCost(int $cost): self
+    {
+        return $this->addValue(LocalParameter::TYPE_COST, $cost);
+    }
+
+    /**
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return $this
+     */
+    public function clearCost(): self
+    {
+        return $this->clearValues(LocalParameter::TYPE_COST);
+    }
+
+    /**
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return array
+     */
+    public function getCost(): array
+    {
+        return $this->getValues(LocalParameter::TYPE_COST);
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function offsetExists($offset): bool
--- a/src/Core/Query/LocalParameters/LocalParameters.php
+++ b/src/Core/Query/LocalParameters/LocalParameters.php
@@ -753,7 +753,7 @@ class LocalParameters implements \ArrayAccess
      */
     public function setCache(bool $cache): self
     {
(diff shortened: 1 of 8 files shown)


src/Core/Query/LocalParameters/LocalParameters.php
```diff
@@ -744,6 +744,70 @@ public function getLocalValues(): array
...
+    /**
+     * @param bool $cache
+     *
+     * @throws \Solarium\Exception\OutOfBoundsException
+     *
+     * @return $this
+     */
+    public function setCache(bool $cache): self
+    {
+        return $this->addValue(LocalParameter::TYPE_CACHE, $cache ? 'true' : 'false');
```
dmaicher: i think we need to clear values here first?

Otherwise calling `->setCache(true)` and then `->setCache(false)` will result in `{!cache=true,false}`? or maybe I misunderstood how it works :smile:

src/QueryType/Select/Query/FilterQuery.php
```diff
@@ -136,6 +136,58 @@ public function setTags(array $tags): self
...
+    /**
+     * Get the information if the filter query should be cached or not.
+     *
+     * @return bool
+     */
+    public function getCache(): bool
+    {
+        $cache = $this->getLocalParameters()->getCache();
+        // The default is to cache the filter Query.
+        return ([] === $cache || 'true' === key($cache)) ? true : false;
```
dmaicher: will `'true' === key($cache)` ever be true? I thought `->getValues(...)` will return the keys only?

https://github.com/solariumphp/solarium/blob/6a7aad5227e1462cd220e0922f155898e75ca73c/src/Core/Query/LocalParameters/LocalParameter.php#L112

dmaicher: See https://github.com/solariumphp/solarium/pull/746/files#diff-3c3f87c60f9b027de6678593757504a6R170

src/Core/Query/LocalParameters/LocalParameters.php
mkalkbrenner: yes, your're right

src/QueryType/Select/Query/FilterQuery.php
```diff
@@ -136,6 +136,58 @@ public function setTags(array $tags): self
...
+    /**
+     * Get the information if the filter query should be cached or not.
+     *
+     * @return bool
+     */
+    public function getCache(): bool
+    {
+        $cache = $this->getLocalParameters()->getCache();
+        // The default is to cache the filter Query.
+        return [] === $cache | 'true' === $cache[0];
```
dmaicher: needs to use `||` or it will fail for empty arrays

mkalkbrenner: I had to push without local testing because I'm currently upgrading PHP.

src/QueryType/Select/Query/FilterQuery.php
```diff
@@ -136,6 +136,58 @@ public function setTags(array $tags): self
...
+    /**
+     * Get the cost of the filter query to be cached or not.
+     *
+     * @return int
+     */
+    public function getCost(): int
+    {
+        $cost = $this->getLocalParameters()->getCost();
+        // The default is to cache the filter Query.
+        return [] === $cost ? 0 : (int) reset($cost);
```
dmaicher: could also be `return (int) reset($cost);`

src/QueryType/Select/Query/FilterQuery.php
mkalkbrenner: I implemented it differently now.

src/QueryType/Select/Query/FilterQuery.php
```diff
@@ -136,6 +136,58 @@ public function setTags(array $tags): self
...
+
+    /**
+     * Get the cost of the filter query to be cached or not.
+     *
+     * @return int
+     */
+    public function getCost(): int
+    {
+        $cost = $this->getLocalParameters()->getCost();
+        // The default is to cache the filter Query.
```
dmaicher: this comment is wrong :smile:

src/QueryType/Select/Query/FilterQuery.php
mkalkbrenner: One error left :-(

mkalkbrenner:
@dmaicher I think I can merge now. Do you agree?

dmaicher:
@mkalkbrenner tested it and it works fine :blush: When will a new release be tagged?

PR #750, opened 2020-02-06
Create and configure Managed Resources

thomascorthals:
I've added commands to create a Managed Resource and configure its initArgs through the REST API. Tests have been updated to include the new functionality.


thomascorthals:
I had to add a Remove command for the managed resources that I created. This would fail with a wrong count if the created resource lingered:

https://github.com/solariumphp/solarium/blob/3282f53462bc8d5255835050dad373e2966ad5cd/tests/Integration/AbstractCoreTest.php#L443-L449

And repeating for different adapters would result in an error.

Running the solr_no_cloud group against my local Solr passes. The Travis CI jobs with SOLR_CLOUD=false fail though. Any pointers on what I'm doing wrong would be greatly appreciated.

thomascorthals:
Sorry for the premature review request. I did figure it out after all. I only ran the Curl tests locally, the errors are caused by the Http adapter not supporting the DELETE method.

mkalkbrenner:
See https://github.com/solariumphp/solarium/pull/738#issuecomment-582568032

Having all these adapters was always a pain.
I would agree to skip this new test for the HTTP adapter and to deprecate the HTTP adapter and the Zend / laminas adapter in 5.2.0.

@wickedOne @thePanz @timohund @jsteggink what's your opinion about it?

wickedOne:
i agree moving to httplug is the way to go. nevertheless, this issue appears to be unrelated to either adapter. 

to replicate you can download yourself a new solr tar, unpack it, fire up the techproducts core and pick ``Techproducts[MyfavouriteAdapter]Test`` and run it three times in a row; this way i managed fail each adapter. this also explains why the http & zend adapter failed because they're the 3rd and 4th in line.

this is the list of managed resources during a 2nd iteration of running tests (breakpoint @ ``\Solarium\Tests\Integration\query::testManagedStopwordsCreation``):
Screenshot 2020-02-10 at 16 47 49
so something is not cleaning up properly which indicates a bug (still makes me wonder why it appears to succeed the first iteration). the ``testCanReloadCore`` causes settings to be reloaded and the managed resources to be-remanaged?
https://lucene.apache.org/solr/guide/8_4/managed-resources.html#applying-managed-resource-changes

apart from that i'm not entirely sure what the ``testManagedResources`` test is supposed to test: count the amount of managed resources provided by the solr demo core? what if that data changes, that wouldn't per sé make the solarium code invalid but tests will start failing; perhaps it's better to test for presence or absence of specific managed resources and preferably ones that are independent of what solr provides?

mkalkbrenner:
Good points.

BTW what about starting the integration test  by creating a core with a random name using the techproducts config-set?

wickedOne:
yes, sounds like a good idea to start tests with a core of which you're sure it's not having artifacts of previous tests. perhaps an [``UNLOAD``](https://lucene.apache.org/solr/guide/8_4/coreadmin-api.html#coreadmin-unload) with a ``deleteInstanceDir=true`` would be easier to manage?

thomascorthals:
I have done some further tests on my local Solr and came to the same conclusion as @wickedOne (before reading any of the comments). It appears only Managed Stopwords are affected by this problem, Managed Synonyms keep working fine no matter how many iterations I ran.

thomascorthals:
@mkalkbrenner I tried starting the test for each adapter with a fresh core. It works—sort of. The problem remains because the configset is shared between the cores. Even if the previous cores are all deleted (random, random_a, random_b and random_new), the config isn't properly cleaned up and the third and fourth test will fail.

If you're still interested in it, I can open a separate PR.

wickedOne:
@thomascorthals did you try using reloads in between actions in the ``testManagedStopwordsCreation``? reading the documentation changes will only take effect after reloading the core so the flow used in testManagedStopwordsCreation is not really a real life scenario. nevertheless it's rather weird changes remain queued for two iterations rather than one

thomascorthals:
@wickedOne I tried it with reloading after Create, Config, Add _and_ Remove. Makes no difference. It doesn't with the changes in PR #751 either, where I unload the core and start with a fresh one.

thomascorthals:
@wickedOne On the other hand, it works if you don't reload the core at all. Comment out these two lines and all the tests pass:

https://github.com/solariumphp/solarium/blob/3282f53462bc8d5255835050dad373e2966ad5cd/tests/Integration/AbstractCoreTest.php#L62-L63

wickedOne:
yes, i assume that would work.
unfortunately it's not about getting the tests to work, but it's about getting the code to do what it's expected to do.

when these integration tests suddenly re-manage deleted managed resources after reloading the core, i think a real life scenario does as well and i don't think it's wise to ship a feature that is broken to begin with...

at the moment i can think of a few scenarios:
- the solr api is broken and the DELETE request doesn't work properly and re-submits managed resources after x reloads
- something gets stuck on the solarium / php side which doesn't properly remove the managed resources (i.e. the raw request data gets re-submitted for whatever reason)
- there's a config we're not aware of which is causing this to happen (just like the autoCommit setting can fool you when you do imports / updates)

unfortunately i won't have time until next weekend to have a proper look at this, but if you're having some additional time on your hands, i'd suggest to first rule out the first scenario.

i think that can easily be done by replicating the tests for stopwords completely in curl commands (including reloads). this way you'll bypass the entire solarium library so if the deleted resources get re-managed it should be that either there's a bug on the solr side, or there's something which should be done additionally (e.g. configure something, execute an additional command)

@mkalkbrenner any thoughts / suggestions on this?

thomascorthals:
I tested with `curl` commands against a clean install with techproducts freshly loaded. I only created and deleted resources, no config changes and nothing added to the list/map.

Stopwords:
- create/delete/create/delete/create/delete/… is OK
- create/reload/delete/create/reload/delete/create/reload/delete/… is OK
- create/delete/reload/create results in an ERROR
- create/delete/create/delete/create/delete/reload/create results in an ERROR

Synonyms:
- create/delete/reload/create/delete/reload/create/delete/reload/… is OK

Stopwords + Synonyms:
- create stop/delete stop/create syn/delete syn/reload/create stop/delete stop/create syn/delete syn/reload/create stop results in an ERROR
- create syn/delete syn/create stop/delete stop/reload/create syn/delete syn/create stop results in an ERROR

Stopwords fail after the **first** RELOAD directly following a Stopwords DELETE or the **second** RELOAD directly following a Synonyms DELETE. Reloading after a CREATE doesn't result in failure, even if there was an earlier DELETE. Synonyms never fail. This is consistent with the failures we're seeing in our tests.

This is a Solr issue. Since it's inconsistent between Stopwords and Synonyms, I assume it's a bug. A setting we're missing would reasonably apply to all Managed Resources.

wickedOne:
thanks a bunch for putting your time and effort into this one!

to me it indeed appears to be some weird behaviour on the solr side...

you might want to consider opening an issue on [their tracker](https://issues.apache.org/jira/projects/SOLR/issues), describe the steps you've taken and see whether they have a reasonable explanation or indeed consider it a bug

thomascorthals:
I circumvented the problem by using unique names for creating the Managed Resources instead of reusing the same name. I had to cheat on `testManagedResources` but I don't think that's a big deal?

https://github.com/solariumphp/solarium/blob/7c5428f870ff8ee47ab20927dfce4ffc8d680337/tests/Integration/AbstractCoreTest.php#L443-L451

@wickedOne suggested to test for specific resources independent of what Solr provides. Isn't that what `testManagedStopwordsCreation` and `testManagedSynonymsCreation` are doing now?

The test fails for Solr 6.6.6. for a different reason now. `getRawData` should return a different class for Solr <7. The version can be obtained from `/admin/info/system` but there is no support for this endpoint yet in Solarium. I'm open to suggestions that don't necessarily involve me implementing `QueryType\Info`.

https://github.com/solariumphp/solarium/blob/7c5428f870ff8ee47ab20927dfce4ffc8d680337/src/QueryType/ManagedResources/Query/Synonyms/Command/Create.php#L36

mkalkbrenner:
> you might want to consider opening an issue on their tracker, describe the steps you've taken and see whether they have a reasonable explanation or indeed consider it a bug

This is a good idea anyway! And the issue should be linked here.
In the Solr community it is common to also post the "question" at the mailing list:
https://lucene.apache.org/solr/community.html

wickedOne:
as for the solr 6.6.6 part: everything < solr 7 is deprecated on the solr side so i'd opt for deprecating the support for that as well (which should be a seperate issue). which makes you wonder whether we should add additional features to 5.x and move all features to 6.x from now on.

as for getting the test to work: as said i don't think cheating and having non failing test is the way to go. if we would document / support a feature in order to delete stopwords which actually triggers a bug, that would potentially cause some trouble down stream.

i'm with @mkalkbrenner and am all for getting it through to the apache solr community and get some kind of response to this issue before we move forward

thomascorthals:
> you might want to consider opening an issue on [their tracker](https://issues.apache.org/jira/projects/SOLR/issues), describe the steps you've taken and see whether they have a reasonable explanation or indeed consider it a bug

I've started [a discussion on the mailing list](https://lucene.472066.n3.nabble.com/Re-creating-deleted-Managed-Stopwords-lists-results-in-error-td4455345.html) that went nowhere.

I've created issue [SOLR-14268](https://issues.apache.org/jira/browse/SOLR-14268).

wickedOne:
ugh, those comments aren't really useful indeed :-(

hopefully response to 14268 will be more constructive than that and at the very least give us some insight on what is happening here

Issue #752, opened 2020-02-14
Solarium 6.0


mkalkbrenner:
At the moment we have several PRs that will introduce great new features.

On the other hand we should consider to get rid of our own HTPP Adapters and introduce compatibility to various  PSRs.

And we should consider to oficially drop support for Solr versions that reached EOL - or at least some of them.

We should discuss the targets of solarium 6.0 here.

But in order to quickly accept the pending PRs I just released 5.1.6 and created the 5.x maintenance branch. `master` is 6.x now :-)

mkalkbrenner:
@wickedOne
@thomascorthals
@jsteggink 
@thePanz 
@timohund 
@localheinz 
@dmaicher

Your input (and contribution) will be appreciated :-)

BTW I created a milestone where we could assign issues and PRs.

My plan is to release 6.0.0-alpha.1 quickly based on the pending PRs

But we should have enough time to still deprecate some stuff in 5.1.x and to refactor some stuff in 6.x.

wickedOne:
i like where this is going :-)

my (personal) milestones:
- support for the [``AnalyticsComponent``](https://lucene.apache.org/solr/guide/8_4/analytics.html). as described in #727  that one is already in pr so that's an easy one.
- support for the [``Schema API``](https://lucene.apache.org/solr/guide/8_4/schema-api.html). i've been working on that last thursday and intent to have something pr worthy this weekend.
- deprecation of the various adapters by supporting [PSR-18](https://www.php-fig.org/psr/psr-18/) as described in #550. if you want to deprecate them in 6.x they it needs to be implemented in 5.x though...
- removal of the symfony event dispatcher dependency and only use interfaces as described in [``PSR-14``](https://www.php-fig.org/psr/psr-14/)
- properly implementing a codestyle standard as discussed in #747 and with that probably improve a bit on the entire workflow by (for instance) using issue templates.
- deprecation of the ``Configurable`` [god object](https://en.wikipedia.org/wiki/God_object). i'll create an issue about that outlining my concerns.

probably some more will come to mind during the weekend B-)

mkalkbrenner:
Since https://github.com/solariumphp/solarium/releases/tag/5.0.0 we basically support the V2 API.

Using https://github.com/solariumphp/solarium/blob/master/src/QueryType/Server/Api/Query.php you can already use the Schema API (and any other Solr API) kind of "manually".

If we consider to drop the support for older Solr versions, we could migrate a lot of stuff from V1 to V2.

[...]

PR #754, opened 2020-02-19
add PSR-18 http adapter

dmaicher:
As mentioned in https://github.com/solariumphp/solarium/issues/752 this fixes https://github.com/solariumphp/solarium/issues/550 by implementing a PSR-18 based `AdapterInterface`.

Todos:
- [x] add more test cases
- [x] request http basic authentication headers
- [x] test it for real with a PSR-18 client implementation on one of my apps with solr
- [x] deprecate other adapters
- [ ] deprecate endpoint timeout setting (will be done in a follow-up PR)
- [x] deprecate adapter options on the client
- [x] use new Psr18Adapter in integration tests

I tested it on one of my Symfony work apps and it works nicely together with http plug bundle and nelmio solarium bundle :blush:  Just a DI compiler pass for wiring the new adapter needed.

works

### DIFF
--- a/src/Core/Client/Adapter/Psr18Adapter.php
+++ b/src/Core/Client/Adapter/Psr18Adapter.php
@@ -135,7 +135,7 @@ class Psr18Adapter extends Configurable implements AdapterInterface
             }
 
             if (!empty($authData['username']) && !empty($authData['password'])) {
-                $headers['Authorization'] = [sprintf('Basic %s', base64_encode($authData['username'].':'.$authData['password']))];
+                $headers['Authorization'] = ['Basic '.base64_encode($authData['username'].':'.$authData['password'])];
             }
         }
 
--- a/src/Core/Client/Adapter/Psr18Adapter.php
+++ b/src/Core/Client/Adapter/Psr18Adapter.php
@@ -64,7 +64,7 @@ class Psr18Adapter extends Configurable implements AdapterInterface
             $psr7Request = $psr7Request->withBody($this->streamFactory->createStream($body));
         }
 
-        foreach ($this->getRequestHeaders($request) as $name => $values) {
+        foreach ($this->getRequestHeaders($request, $endpoint) as $name => $values) {
             foreach ($values as $value) {
                 $psr7Request = $psr7Request->withAddedHeader($name, $value);
             }
@@ -108,7 +108,7 @@ class Psr18Adapter extends Configurable implements AdapterInterface
         return $request->getRawData();
     }
 
-    private function getRequestHeaders(Request $request): array
+    private function getRequestHeaders(Request $request, Endpoint $endpoint): array
     {
         $headers = [];
 
@@ -124,6 +124,18 @@ class Psr18Adapter extends Configurable implements AdapterInterface
                 $headers['Content-Type'] = ['application/x-www-form-urlencoded; charset=utf-8'];
             } else {
                 $headers['Content-Type'] = ['application/xml; charset=utf-8'];
+            }
+        }
+
+        if (!isset($headers['Authorization'])) {
+            // Try endpoint authentication first, fallback to request for backwards compatibility
+            $authData = $endpoint->getAuthentication();
+            if (empty($authData['username'])) {
+                $authData = $request->getAuthentication();
+            }
+
+            if (!empty($authData['username']) && !empty($authData['password'])) {
+                $headers['Authorization'] = [sprintf('Basic %s', base64_encode($authData['username'].':'.$authData['password']))];
             }
         }
 
--- a/src/Core/Client/Adapter/Psr18Adapter.php
+++ b/src/Core/Client/Adapter/Psr18Adapter.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace Solarium\Core\Client\Adapter;
+
+use Psr\Http\Client\ClientExceptionInterface;
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestFactoryInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Solarium\Core\Client\Endpoint;
+use Solarium\Core\Client\Request;
+use Solarium\Core\Client\Response;
+use Solarium\Core\Configurable;
+use Solarium\Exception\HttpException;
+
+class Psr18Adapter extends Configurable implements AdapterInterface
+{
+    /**
+     * @var ClientInterface
+     */
+    private $httpClient;
+
+    /**
+     * @var RequestFactoryInterface
+     */
+    private $requestFactory;
+
+    /**
+     * @var StreamFactoryInterface
+     */
+    private $streamFactory;
+
+    public function __construct(
+        ClientInterface $httpClient,
+        RequestFactoryInterface $requestFactory,
+        StreamFactoryInterface $streamFactory
+    ) {
+        parent::__construct();
+        $this->httpClient = $httpClient;
+        $this->requestFactory = $requestFactory;
+        $this->streamFactory = $streamFactory;
+    }
+
+    public function execute(Request $request, Endpoint $endpoint): Response
+    {
+        try {
+            return $this->createResponse($this->httpClient->sendRequest($this->createPsr7Request($request, $endpoint)));
+        } catch (ClientExceptionInterface $clientException) {
+            throw new HttpException("HTTP request failed, {$clientException}");
+        }
+    }
+
+    private function createPsr7Request(Request $request, Endpoint $endpoint): RequestInterface
+    {
+        $uri = AdapterHelper::buildUri($request, $endpoint);
+
+        $psr7Request = $this->requestFactory->createRequest(
+            $request->getMethod(),
+            $uri
+        );
+
+        if (null !== $body = $this->getRequestBody($request)) {
+            $psr7Request = $psr7Request->withBody($this->streamFactory->createStream($body));
+        }
+
+        foreach ($request->getHeaders() as $headerLine) {
+            [$header, $value] = explode(':', $headerLine);
+            if ($header = trim($header)) {
+                $psr7Request = $psr7Request->withAddedHeader($header, $value);
+            }
+        }
+
+        return $psr7Request;
+    }
+
+    private function createResponse(ResponseInterface $psr7Response): Response
+    {
+        $headerLines = [
+            sprintf(
+                'HTTP/%s %s %s',
+                $psr7Response->getProtocolVersion(),
+                $psr7Response->getStatusCode(),
+                $psr7Response->getReasonPhrase()
+            ),
+        ];
+
+        foreach ($psr7Response->getHeaders() as $name => $values) {
+            $headerLines[] = sprintf('%s: %s', $name, implode(', ', $values));
+        }
+
+        return new Response((string) $psr7Response->getBody(), $headerLines);
+    }
+
+    private function getRequestBody(Request $request): ?string
+    {
+        if (Request::METHOD_PUT == $request->getMethod()) {
+            return $request->getRawData();
+        }
+
+        if (Request::METHOD_POST !== $request->getMethod()) {
+            return null;
+        }
+
+        if ($request->getFileUpload()) {
+            return AdapterHelper::buildUploadBodyFromRequest($request);
+        }
+
+        return $request->getRawData();
+    }
+}
--- a/src/Core/Client/Adapter/Psr18Adapter.php
+++ b/src/Core/Client/Adapter/Psr18Adapter.php
@@ -14,7 +14,7 @@ use Solarium\Core\Client\Response;
 use Solarium\Core\Configurable;
 use Solarium\Exception\HttpException;
 
-class Psr18Adapter extends Configurable implements AdapterInterface
+final class Psr18Adapter extends Configurable implements AdapterInterface
 {
     /**
      * @var ClientInterface
--- a/src/Core/Client/Adapter/Psr18Adapter.php
+++ b/src/Core/Client/Adapter/Psr18Adapter.php
@@ -64,10 +64,9 @@ class Psr18Adapter extends Configurable implements AdapterInterface
             $psr7Request = $psr7Request->withBody($this->streamFactory->createStream($body));
         }
 
-        foreach ($request->getHeaders() as $headerLine) {
-            [$header, $value] = explode(':', $headerLine);
-            if ($header = trim($header)) {
-                $psr7Request = $psr7Request->withAddedHeader($header, $value);
+        foreach ($this->getRequestHeaders($request) as $name => $values) {
+            foreach ($values as $value) {
+                $psr7Request = $psr7Request->withAddedHeader($name, $value);
             }
         }
 
@@ -107,5 +106,27 @@ class Psr18Adapter extends Configurable implements AdapterInterface
         }
 
         return $request->getRawData();
+    }
+
+    private function getRequestHeaders(Request $request): array
+    {
+        $headers = [];
+
+        foreach ($request->getHeaders() as $headerLine) {
+            [$header, $value] = explode(':', $headerLine);
+            if ($header = trim($header)) {
+                $headers[$header][] = $value;
+            }
+        }
+
+        if (!isset($headers['Content-Type'])) {
+            if (Request::METHOD_GET == $request->getMethod()) {
+                $headers['Content-Type'] = ['application/x-www-form-urlencoded; charset=utf-8'];
+            } else {
+                $headers['Content-Type'] = ['application/xml; charset=utf-8'];
+            }
+        }
+
+        return $headers;
     }
 }
--- a/src/Core/Client/Client.php
+++ b/src/Core/Client/Client.php
@@ -5,6 +5,7 @@ namespace Solarium\Core\Client;
 use Solarium\Core\Client\Adapter\AdapterInterface;
 use Solarium\Core\Client\Adapter\Curl;
 use Solarium\Core\Configurable;
+use Solarium\Core\ConfigurableInterface;
 use Solarium\Core\Event\Events;
 use Solarium\Core\Event\PostCreateQuery as PostCreateQueryEvent;
 use Solarium\Core\Event\PostCreateRequest as PostCreateRequestEvent;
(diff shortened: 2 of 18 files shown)


dmaicher:
@wickedOne @mkalkbrenner what should we do about the timeout setting of endpoints?

This is impossible to support on a per endpoint/request basis with psr-7. The http client itself will have to be configured with a timeout.

So this will not work anymore:
https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Adapter/Guzzle.php#L47

Should we deprecate the timeout setting completely in 5.x and remove it in version 6? Or what do you think?

wickedOne:
can't that be achieved by passing a [``Request-Timeout``](https://tools.ietf.org/id/draft-thomson-hybi-http-timeout-00.html#rfc.section.1.1) header using the ``MessageInterface#withHeader`` method (ideally with a ``hasHeader`` check)?

i think deprecating the timeout setting completely will break quite a bit especially because it's currently done on an endpoint base (i.e. migration won't be as simple as moving a setting from solarium to your http client)

dmaicher:
@wickedOne I don't think the `Request-Timeout` can be used to achieve the same.

So as far as I understand the `Request-Timeout` header will be used server-side:

> An origin server interprets this header as the time between receipt of a complete request and the time that it generates and begins sending the response.

Also it really seems a bit useless:
>  An origin server MAY choose to take longer to produce a response, at the risk that the client is no longer able to use the response

Some scenarios that currently with the Guzzle and Curl adapters will timeout:
- network issues that result in hanging connections and the origin server does not even receive the request
- origin server accepted the request but just takes really long to respond

So I don't see how to "solve" those with request headers :confused: The http client itself will have to be configured with such timeouts.

wickedOne:
true, it won't solve any client side timeouts but those shouldn't be solarium's concern.

apparently solr itself [relies on the config of it's servlet container](https://cwiki.apache.org/confluence/display/solr/FAQ#FAQ-Whydoestherequesttimeoutsometimeswhendoingcommits) to determine timeouts. 
even using the [``timeAllowed``](https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html#CommonQueryParameters-ThetimeAllowedParameter) parameter potentially returns partial results which probably causes more problems than it solves...

in short, telling solr to timeout is of no use (should have done a bit more digging before responding to your initial question, sorry for that).

looking at the commit history of that option, it has been there since the beginning and no test explicitly states a use on the solr side of things.

seems like deprecating this option is the way to go after all

dmaicher:
@wickedOne another thing: Since we are deprecating other adapters this will be an issue...

https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Client.php#L184

What should we do about this `adapter` option? Seems we also need to deprecate it. We cannot provide any default adapter instance anymore in Solarium 6 since there is no default psr-18 implementation? 

Also we should then deprecate `adapteroptions`:
https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Client.php#L524

And the `AdapterInterface` should ideally not extend `ConfigurableInterface` anymore in Solarium 6.

wickedOne:
agreed, both the ``adapteroptions`` and ``adapter`` option should be deprecated.

i guess your ``Psr18Adapter`` will be the new default adapter (ideally as client constructor argument) and when you simplify the following (while probably deprecating the later):

```php
    public function setAdapter(AdapterInterface $adapter): ClientInterface
    {
        $this->adapter = $adapter;

        return $this;
    } 

// ... 

    protected function createAdapter():void
    {
    }
```

solarium implementations can still chose to use their custom adapter and you indeed can remove the ``ConfigurableInterface`` extend as well

dmaicher:
@wickedOne I don't see any deprecation path for making `AdapterInterface` not extend `ConfigurableInterface` :confused: 

https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Adapter/AdapterInterface.php#L25

Any ideas? I guess this can only be done in 6.0 then without any triggered deprecations in 5.x?

wickedOne:
i see you're making progress: awesome!

as for the extend deprecation: one way to do it might be:
```php
public function setAdapter(AdapterInterface $adapter): self
{
    if (true === is_subclass_of($adapter, ConfigurableInterface:class)) {
        @trigger_error(sprintf('Configuring an adapter through the %s class is deprecated in solarium 5.x and will be removed in solarium 6.x. Use fully configured adapters only', ConfigurableInterface:class), E_USER_DEPRECATED);
    }

    $this->adapter = $adapter;

    return $this;
}
```

i think the ``setAdapter`` method is the only one which accepts an ``AdapterInterface`` from outside the library so deprecating the extend in the setter should be sufficient

dmaicher:
@wickedOne the problem with that is that people cannot get rid of this deprecation while still using Solarium 5.x as the `AdapterInterface` will still extend `ConfigurableInterface` :confused: 

So its a deprecation that you cannot fix until you actually move to version 6. Not sure that makes sense?

wickedOne:
agreed, it's not the most ideal situation as people can't get rid of the deprecation themselves.

nevertheless they get a heads up that the implementation is about to change, giving them time to adapt if necessary.

mkalkbrenner:
Regarding timeouts

> true, it won't solve any client side timeouts but those shouldn't be solarium's concern.
> 
> apparently solr itself [relies on the config of it's servlet container](https://cwiki.apache.org/confluence/display/solr/FAQ#FAQ-Whydoestherequesttimeoutsometimeswhendoingcommits) to determine timeouts.
> even using the [`timeAllowed`](https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html#CommonQueryParameters-ThetimeAllowedParameter) parameter potentially returns partial results which probably causes more problems than it solves...
> 
> in short, telling solr to timeout is of no use (should have done a bit more digging before responding to your initial question, sorry for that).
> 
> looking at the commit history of that option, it has been there since the beginning and no test explicitly states a use on the solr side of things.
> 
> seems like deprecating this option is the way to go after all

I agree that  the timeout setting should not be solarium's concern anymore. The application has to deal with the HTTP implementation directly.

But for your information:
In search API Solr we distinguish between four timeout settings:
- query timeout
- index timeout
- optimize timout
- finalize timeout

All of them are needed, especially background tasks to index content or to trigger expensive streaming expressions require a higher non-default timeout.

But due to the fact that we already work around the single timeout setting in solarium, we could easily migrate that code to manipulate the "HTTP service provider" directly.

A second difficulty comes into my mind. Solarium offers a plugin that allows parallel requests. I think this one requires curl. But I think that it is acceptable to have plugins that don't work with all http providers.

dmaicher:
@mkalkbrenner ok I was not aware of this plugin :cry: 

https://github.com/solariumphp/solarium/blob/master/src/Plugin/ParallelExecution/ParallelExecution.php#L107

What should we do about this then? The plan was to deprecate the CurlAdapter and remove it in 6.0 :confused: 

Or we keep the Curl Adapter and only deprecate/remove all the others then...

wickedOne:
@mkalkbrenner do you know a way to explicitly set those timeouts when requesting solr?

if not we might want to consider catching a ``NetworkExceptionInterface`` and throw an appropriate exception?

as far as i know psr-18 doesn't dictate the protocol so it would be impossible for us to determine whether curl is used or not.

mkalkbrenner:
> Or we keep the Curl Adapter and only deprecate/remove all the others then...

I think that this is an acceptable trade-off. The curl adapter uses plain PHP while the frameworks should be "hidden" by PSR18.
Following this argument we can also consider keeping the Http Adapter.

Even if we keep these two pure php adapters, the PSR18 adapter gives us the benefit to not be responsible to follow the development of the different frameworks and it adds support for new frameworks we don't support yet.

So with 6.0.0 we'll have three adapters:
- Http
- Curl
- Psr18

> i guess your Psr18Adapter will be the new default adapter 

If we keep the native ones we should consider to keep Curl as default.

> @mkalkbrenner do you know a way to explicitly set those timeouts when requesting solr?

In search_api_solr we use the default curl adapter and this one does
```php
        curl_setopt($handler, CURLOPT_TIMEOUT, $options['timeout']);
        curl_setopt($handler, CURLOPT_CONNECTTIMEOUT, $options['timeout']);
```

wickedOne:
so those values are client side and are not passed to solr.
i think we should ask ourselves whether solarium should be the one being responsible for that.

is it a real life scenario to be able to configure that option on a per request basis? if so, would it be acceptable to typehint a PSR-7 ``RequestInterface``  whenever Solarium's client passes around a request and have people modify it's options if they want to using the post create request event?

thomascorthals:
> is it a real life scenario to be able to configure that option on a per request basis?

You can execute requests against an additional endpoint with a different timeout option. That's functionally equivalent to a timeout option per request, isn't it?

mkalkbrenner:
I wanted to merge the PR right now. But I noticed that we didn't yet agree on https://github.com/solariumphp/solarium/pull/754#issuecomment-595698473

@thomascorthals From my point view the only thing missing to merge it to 5.x and to have the deprecations in 5.2 is to remove the deprecations on the HTTP and the Curl Adapters - if you agree!

src/Core/Client/Client.php
```diff
@@ -279,11 +280,31 @@ class Client extends Configurable implements ClientInterface
...
      * If an EventDispatcher instance is provided this will be used instead of creating a new instance
      *
-     * @param array                    $options
-     * @param EventDispatcherInterface $eventDispatcher
+     * @param AdapterInterface              $adapter
+     * @param array|null                    $options
+     * @param EventDispatcherInterface|null $eventDispatcher
      */
-    public function __construct(array $options = null, EventDispatcherInterface $eventDispatcher = null)
+    public function __construct($adapter = null, $options = null, $eventDispatcher = null)
```
dmaicher: Should we allow passing null then for `$adapter` and fall back to instantiate a `Curl` adapter?

Should the constructor signature then be

```php
public function __construct(array $options = null, EventDispatcherInterface $eventDispatcher = null, AdapterInterface $adapter = null)
```

thomascorthals:
@mkalkbrenner I agree that `Curl` can't be deprecated if it's required for `ParallelExecution`. I'm indifferent to the `Http` adapter, but I see the logic in only deprecating framework specific adapters.

Keeping `Curl` as the default for 6.0.0 will give us some time to test drive `Psr18` first.

src/Core/Client/Client.php
thomascorthals: If the constructor signature changes, a new optional argument is best added as last. Making it the first argument in the new signature would break backward compatibility.

mkalkbrenner: I think we should keep the "dualism" @dmircher introduced in the code below. The first parameter could be an adapter instance or an options array. We already use such a pattern in the code base. But we should name the parameter  something like $adapter_or_options and the type needs to be AdapterInterface|array|null

On null we should keep the existing behaviour and use the curl adapter.

wickedOne: this can turn ugly very quickly, but i'm with @mkalkbrenner and @dmaicher to have the ``$adapter`` parameter as the first argument...

not entirely sure whether it would be added value to rename the parameter to something different and in the end change it back to ``$adapter`` as this is supposed to be a temporary solution. i'd opt for a user deprecation error in the bit below where ``if (is_array($adapter)) {`` returns true.

dmaicher: So I kept the adapter as the first argument. In Solarium 6 it should be this signature then:

```php
public function __construct(AdapterInterface $adapter = null, array $options = null, EventDispatcherInterface $eventDispatcher = null)

File: tests/Integration/TestClientFactory.php [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+
+use Nyholm\Psr7\Factory\Psr17Factory;
+use Solarium\Client;
+use Http\Adapter\Guzzle6\Client as GuzzlePsrClient;
+use Solarium\Core\Client\Adapter\Psr18Adapter;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+final class TestClientFactory
+{
+    public static function createWithPsr18Adapter(array $options = null, EventDispatcherInterface $eventDispatcher = null): Client
```
dmaicher: shall we keep this? Or should we just use the default `Curl` adapter for those tests like before?

src/Core/Client/Endpoint.php
```diff
@@ -194,9 +194,13 @@ public function getCore(): ?string
      * @param int $timeout
      *
      * @return self Provides fluent interface
+     *
+     * @deprecated Endpoint::setTimeout is deprecated since Solarium 5.2 and will be removed in Solarium 6. Configure the timeout on the HTTP Client used by the Adapter instead.
      */
     public function setTimeout(int $timeout): self
     {
+        @trigger_error('Endpoint::setTimeout is deprecated since Solarium 5.2 and will be removed in Solarium 6. Configure the timeout on the HTTP Client used by the Adapter instead.', E_USER_DEPRECATED);
```
dmaicher: Shall we keep this deprecation then?

The timeout is currently used in the non-deprecated Adapters `Curl` and `Http`... so for those there is no upgrade path yet

dmaicher:
@mkalkbrenner seems travis builds are not triggered anymore? (maybe because of the 5.x base branch?)

mkalkbrenner:
@dmaicher it worked a few hours ago when I created the PR for tests against Solr 8.5.0.
But as the travis user inferface doesn't load at the moment it looks like an issue on travis-ci.org.

src/Core/Client/Client.php
wickedOne: i'm a bit on the fence about this one...

i don't think the ``AdapterInterface`` should be nullable as the client can't function without it and i'm not entirely sure falling back to a default should be the task of the client itself.

same story for the ``EventDispatcherInterface``; all essential methods (``createRequest``, ``createResult``, ``execute`` & ``executeRequest``) in the client expect the event dispatcher not to be null. to add insult to injury, solarium listens to itself: the ``PostBigRequest`` won't function with the absence of an event dispatcher. once more i don't think defining the dispatcher should be tasked to the client.

which leaves us with the ``$options`` parameter being the only optional one.
so personally i'd prefer the following signature:

```php
public function __construct(AdapterInterface $adapter, EventDispatcherInterface $eventDispatcher, array $options = [])
```

this would require some additional deprecation warnings and of course has a bit more of an impact, but if we're changing constructor signatures i think a discussion about it prevents us from doing it more than once.

so, any thoughts?

mkalkbrenner: @wickedOne I see your point. But I think my suggestion in https://github.com/solariumphp/solarium/pull/754#discussion_r399118932 is a good intermediate step for 5.2 as it is fully compatible for existing installations and allows us to fire deprecation warnings.
For 6.0 we can continue to discuss the API change as 5.2 opens the window to announce it.

dmaicher: Ok I agree here. So now the constructor should be fully backwards compatible with deprecations.

In Solarium 6 we can change it then with proper type hints.

Note: I did not change the nullable event dispatcher argument. I think this should be done in the scope of some other PR before we tag a new release for 5.2 then.

wickedOne: @mkalkbrenner yes i agree with your comment and the solution provided by @dmaicher 
my previous comment in this thread was about the final signature for solarium 6.x

changes in this pr should enable us to add appropriate deprecation warnings for the final signare as we're checking anyway. 

so i would suggest the following additional warning:
```php
if (is_array($adapter)) {
    @trigger_error('passing options as the first constructor argument is deprecated in solarium 5.2 and will be removed in 6.x. use the third argument instead', E_USER_DEPRECATED);
    // ...
```

src/Core/Client/Endpoint.php
mkalkbrenner: I suggest to keep the deprecation here. It will be kind of inconvienient for users of the curl of adapter. Therefore we should mention in the documentation how to port your code to solarium 6 (not now).

tests/Integration/TestClientFactory.php
mkalkbrenner: For Solrium 5.2 we should fall back to curl.

src/Core/Client/Adapter/Psr18Adapter.php
```diff
@@ -0,0 +1,144 @@
...
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Solarium\Core\Client\Endpoint;
+use Solarium\Core\Client\Request;
+use Solarium\Core\Client\Response;
+use Solarium\Core\Configurable;
+use Solarium\Exception\HttpException;
+
+final class Psr18Adapter extends Configurable implements AdapterInterface
```
mkalkbrenner: What is the reason for "final" (or the advantage)?

dmaicher: I think we should not open this class for extension. Its does not make sense anyway as everything is private except `public function execute(Request $request, Endpoint $endpoint): Response`.

So we should not even encourage people to extend it. They should use composition if they need to customize stuff.

Or what do you think?

src/Core/Client/Endpoint.php
dmaicher: But on Solarium 5.2 I cannot get rid of this deprecation which is not good. We should already have an alternative then.

src/Core/Client/Client.php
dmaicher: @wickedOne I know changed it also for the event dispatcher.

Please check it again. I also added tests for the triggered deprecations using the Symfony phpunit bridge.

wickedOne: @dmaicher yes, looks good to me!

still think an additional user deprecation when passing the options argument as the first argument is a good think to do as then it provides all the necessary warnings for moving from 5.2 to 6.0:
```php
        if (is_array($adapter)) {
            @trigger_error('passing options as the first constructor argument is deprecated in solarium 5.2 and will be removed in 6.0. use the third argument instead', E_USER_DEPRECATED);

            $options = $adapter;
        }
```
what do you think?

ping @mkalkbrenner

dmaicher: But then they will get two deprecations. I think this one is clear enough?

https://github.com/solariumphp/solarium/pull/754/files#diff-f09ef20cbac899d71549b17bb68ceb2cR300

wickedOne: yes, you're right

src/Core/Client/Endpoint.php
mkalkbrenner: You're right.

In the future people would have to adjust timeouts directly in the component they inject, for example guzzle.

Let's follow this path. I suggest to copy the timeout getter an setter to the curl and http themselves. In the Client interface we already have getAdapter(). So if you use curl or http adapters you can get rid of the deprecation by calling `$client->getAdapter()->setTimeout()`.

src/Core/Client/Adapter/Psr18Adapter.php
mkalkbrenner: OK. Let's leave it like this and wait until someone complains ;-)

src/Core/Client/Endpoint.php
dmaicher: @mkalkbrenner I propose to do the endpoint timeout deprecation in a follow-up PR. I will work on it once this one is approved :wink:

dmaicher:
@wickedOne @thomascorthals @mkalkbrenner ready for another round of reviews :wink: 

the endpoint timeout deprecation (and making it configurable on the `Curl` and `Http` Adapters will be done in a follow-up PR)

Issue #755, opened 2020-02-21
CustomizeRequest doesn't work well with PostBigRequest


efpapado:
Hi,
I'm trying to use both CustomizeRequest and PostBigRequest for a request, but I can't get them to work together.

While building the query, I'm adding a couple of custom parameters. Then I'm converting the request to POST because the request is getting too large. But the custom parameters are not added to the POST rawData.

I followed the code until `\Solarium\Core\Client\Client::createRequest` and after the `POST_CREATE_REQUEST` is dispatched I can see that rawData only contains the parameters that were added through the Query object, not the ones that were added to the Client through CustomizeRequest.

Is it a bug, or do I miss something?

Edit: I'm using the 5.1.5 version of the library.

wickedOne:
it looks like the ``CustomizeRequest`` hooks into the ``Events::PRE_EXECUTE_REQUEST`` while you want it to hook into the ``Events::PRE_CREATE_REQUEST`` so the ``PostBigRequest`` takes the custimizations into consideration while converting from GET to POST.

indeed looks like to a bug to me

efpapado:
A bit more info on the debugging, based on what I found here: [https://solarium.readthedocs.io/en/stable/customizing-solarium/](https://solarium.readthedocs.io/en/stable/customizing-solarium/)
it seems that the PostBigRequest is handled on the POST_CREATE_REQUEST event, while the CustomizeRequest on the PRE_EXECUTE_REQUEST which comes after.
So it's not possible to convert the parameters that are added by CustomizeRequest to POST rawData, because the GET-->POST conversion is happening too early.

efpapado:
Haha @wickedOne  we posted the same info at the same time :D

wickedOne:
haha @efpapado , what are the odds? :-)

if you need to get around this issue quickly, you might want to use the ``registerPlugin`` method of the Client class to register your own ``CustomizeRequest`` plugin which hooks into the appropriate event. 

and of course feel free to open a pull request to get it fixed in the solarium code base

thomascorthals:
Hooking `CustomizeRequest` into `Events::PRE_CREATE_REQUEST` will also fix #350.

efpapado:
I'll try to create a PR on Monday that I'm back at work, thanks for the help :)

efpapado:
[...]

PR #756, opened 2020-02-24
Refactor PostBigRequest plugin to attach on PRE_EXECUTE_REQUEST event instead of POST_CREATE_REQUEST.

efpapado:


efpapado:
Fixes solariumphp/solarium#755 (https://github.com/solariumphp/solarium/issues/755)

wickedOne:
thanks @efpapado !

i see your concern regarding the ``PRE_CREATE_REQUEST`` event and agree that modifying the ``RequestBuilderInterface`` would be too much of a change in order to get this issue resolved.

however, moving the modification of the request (as happens in the ``PostBigRequest``) outside of the request event chain, possibly breaks implementations of the ``ParallelExecution`` plugin so  might not be the way to go either.

as solarium has a dependency on the Symfony EventDispatcher we might consider having both the ``PostBigRequest`` and ``CustomizeRequest`` hook into the ``Events::POST_CREATE_REQUEST`` and make sure the one from the ``CustomizeRequest`` is executed **first** by [passing a priority](https://symfony.com/doc/current/components/event_dispatcher.html#connecting-listeners) as a third parameter while registering the listeners.

and could you add a unit test confirming that these modifications do actually solve the problem as described in #755 ?

PR #757, opened 2020-02-24
Added raw XML commands to Update query

thomascorthals:
I've added the ability to add raw XML commands directly to an `Update` query. This will facilitate importing the techproducts example XMLs for #751.

Commands wrapped in `<update>...</update>` have those tags stripped because we always add them in the `RequestBuilder`. This is naïvely done under the assumption that the provided XML is valid and doesn't contain any edge cases (e.g. `<update>` inside a comment or a processing instruction). Insignificant whitespace and innocent comments don't throw it off. I would have to actually parse the XML to cover all possibilities, but that would introduce a dependency on an XML parser for something that might have little usage outside our own integration tests.

There is no integration test yet, #751 will include that.

I have documented it, although users shouldn't be building XML strings in their own code. I think I did get that point across.

### DIFF
--- a/src/QueryType/Update/Query/Command/Raw.php
+++ b/src/QueryType/Update/Query/Command/Raw.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Solarium\QueryType\Update\Query\Command;
+
+use Solarium\QueryType\Update\Query\Query as UpdateQuery;
+
+/**
+ * Update query raw XML command.
+ *
+ * @see http://wiki.apache.org/solr/UpdateXmlMessages
+ */
+class Raw extends AbstractCommand
+{
+    /**
+     * XML command strings to send.
+     *
+     * @var array
+     */
+    protected $commands = [];
+
+    /**
+     * Get command type.
+     *
+     * @return string
+     */
+    public function getType(): string
+    {
+        return UpdateQuery::COMMAND_RAW;
+    }
+
+    /**
+     * Add a single XML command string to the command.
+     *
+     * @param string $command
+     *
+     * @return self Provides fluent interface
+     */
+    public function addCommand($command): self
+    {
+        $this->commands[] = $command;
+
+        return $this;
+    }
+
+    /**
+     * Add multiple XML command strings to the command.
+     *
+     * @param array $commands
+     *
+     * @return self Provides fluent interface
+     */
+    public function addCommands(array $commands): self
+    {
+        $this->commands = array_merge($this->commands, $commands);
+
+        return $this;
+    }
+
+    /**
+     * Get all XML command strings of this command.
+     *
+     * @return array
+     */
+    public function getCommands(): array
+    {
+        return $this->commands;
+    }
+
+    /**
+     * Build XML command strings based on options.
+     */
+    protected function init()
+    {
+        $command = $this->getOption('command');
+        if (null !== $command) {
+            if (is_array($command)) {
+                $this->addCommands($command);
+            } else {
+                $this->addCommand($command);
+            }
+        }
+    }
+}
--- a/src/QueryType/Update/Query/Command/RawXML.php
+++ b/src/QueryType/Update/Query/Command/RawXML.php
@@ -9,7 +9,7 @@ use Solarium\QueryType\Update\Query\Query as UpdateQuery;
  *
  * @see http://wiki.apache.org/solr/UpdateXmlMessages
  */
-class Raw extends AbstractCommand
+class RawXML extends AbstractCommand
 {
     /**
      * XML command strings to send.
@@ -25,7 +25,7 @@ class Raw extends AbstractCommand
      */
     public function getType(): string
     {
-        return UpdateQuery::COMMAND_RAW;
+        return UpdateQuery::COMMAND_RAWXML;
     }
 
     /**
@@ -35,7 +35,7 @@ class Raw extends AbstractCommand
      *
      * @return self Provides fluent interface
      */
-    public function addCommand($command): self
+    public function addCommand(string $command): self
     {
         $this->commands[] = $command;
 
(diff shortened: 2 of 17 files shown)


src/QueryType/Update/Query/Command/Raw.php
```diff
@@ -0,0 +1,83 @@
...
+namespace Solarium\QueryType\Update\Query\Command;
+
+use Solarium\QueryType\Update\Query\Query as UpdateQuery;
+
+/**
+ * Update query raw XML command.
+ *
+ * @see http://wiki.apache.org/solr/UpdateXmlMessages
+ */
+class Raw extends AbstractCommand
```
mkalkbrenner: As mentioned above the class should be named explicit `RawXML` as there're raw JSON commands, too.

src/QueryType/Update/Query/Command/Raw.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Add a single XML command string to the command.
+     *
+     * @param string $command
+     *
+     * @return self Provides fluent interface
+     */
+    public function addCommand($command): self
```
mkalkbrenner: The parameter type hint should be strictly `string` unless we support other types like \DOMDocument

src/QueryType/Update/Query/Query.php
```diff
@@ -44,14 +45,19 @@ class Query extends BaseQuery
     const COMMAND_COMMIT = 'commit';
 
     /**
-     * Update command rollback.
+     * Update command optimize.
      */
-    const COMMAND_ROLLBACK = 'rollback';
+    const COMMAND_OPTIMIZE = 'optimize';
```
mkalkbrenner: why did you delete 'rollback'?

src/QueryType/Update/Query/Query.php
```diff
@@ -44,14 +45,19 @@ class Query extends BaseQuery
...
 
     /**
      * Update command optimize.
      */
-    const COMMAND_OPTIMIZE = 'optimize';
+    const COMMAND_RAW = 'raw';
```
mkalkbrenner: COMMAND_RAWXML

src/QueryType/Update/Query/Query.php
mkalkbrenner: I see, you changed the order.

thomascorthals: What would be the best order? There is no single correct 'natural flow' for these commands. Should I change them to alphabetical order? The order used in `RequestBuilder`'s `switch/case`?

src/QueryType/Update/Query/Command/Raw.php
thomascorthals: I just forgot the type hint.

I don't think supporting `\DOMDocument` is necessary. Our intended use is reading existing XML from a file. If anyone wants to build the XML outside of Solarium, they can use that library's way of getting the string. That avoids introducing a dependency, albeit on a library that is enabled by default.

src/QueryType/Update/Query/Query.php
thomascorthals: I went with alphabetical. That's more or less consistent with other `const`s in the codebase.

thomascorthals:
I've refactored Raw to RawXML. This PR is a prerequisite for importing the example data in #751. Once that's done, we'll have an integration test.

src/QueryType/Update/Query/Command/Raw.php
thomascorthals: Besides, if it took a `\DOMDocument`, it wouldn't be "raw".

src/QueryType/Update/Query/Command/RawXML.php
```diff
@@ -0,0 +1,83 @@
...
+namespace Solarium\QueryType\Update\Query\Command;
+
+use Solarium\QueryType\Update\Query\Query as UpdateQuery;
+
+/**
+ * Update query raw XML command.
+ *
+ * @see http://wiki.apache.org/solr/UpdateXmlMessages
+ */
+class RawXML extends AbstractCommand
```
mkalkbrenner: Just for consistency we should camel case XML => RawXml

src/QueryType/Update/RequestBuilder.php
```diff
@@ -181,6 +184,30 @@ public function buildRollbackXml(): string
...
+     *
+     * @return string
+     */
+    public function buildRawXmlXml(RawXML $command): string
+    {
+        $xml = '';
+
+        foreach ($command->getCommands() as $raw) {
+            // unwrap grouped commands, they must be consolidated in a single <update>
+            if (false !== ($pos = strpos($raw, '<update'))) {
```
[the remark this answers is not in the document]
thomascorthals: > Doesn't it make sense to immediately "unwrap" the command in RawXml::addCommand()?

That would mean, from a user's perspective, that `getCommands` doesn't return the 'same' commands that were added with `addCommand` or `addCommands`. I think it would be fair to expect `assertSame` to work on add/get.

Furthermore, I'm already thinking about how to go about raw JSON commands. The way we build requests, it will be straightforward to generate XML from the decoded JSON string. I wouldn't do that in `addCommand` because a user will definitely expect JSON back from `getCommands`. It would break the symmetry if you could get your verbatim JSON string back but not your verbatim XML string.

> It could be done using a simple multi line preg_replace().

I tried and failed. My current code works for multiple comments and/or PIs before *and* after the root element, with or without whitespace before, after and inbetween those. It also works for whitespace and attributes in the opening tag. (Attributes not throwing it off is incidental. There are no documented attributes, and Solr happily ignores them too.) I couldn't get that to work for every valid combination if I wanted to make sure only a root `<update>` element gets picked up. If we can safely assume there will never ever be a nested `<update>` element anywhere in a valid XML command, it could be done with little effort. Do we feel confident about that assumption?

src/QueryType/Update/Query/Command/RawXML.php
thomascorthals: 🐫!

mkalkbrenner:
I'm still not sure about the "magic" to deal with the update element in XML. But we can discuss this further in the 6.0 development cycle and I shouldn't block the PR into master :-)

PR #758, opened 2020-02-26
overwrite can be null

wickedOne:
fixed too strict typehint in the buffered add pre commit event

### DIFF
--- a/src/Plugin/BufferedAdd/BufferedAdd.php
+++ b/src/Plugin/BufferedAdd/BufferedAdd.php
@@ -265,7 +265,7 @@ class BufferedAdd extends AbstractPlugin
         $event = new PreCommitEvent($this->buffer, $overwrite, $softCommit, $waitSearcher, $expungeDeletes);
         $this->client->getEventDispatcher()->dispatch($event, Events::PRE_COMMIT);
 
-        $this->updateQuery->addDocuments($this->buffer, $event->getOverwrite());
+        $this->updateQuery->addDocuments($event->getBuffer(), $event->getOverwrite());
         $this->updateQuery->addCommit($event->getSoftCommit(), $event->getWaitSearcher(), $event->getExpungeDeletes());
         $result = $this->client->update($this->updateQuery, $this->getEndpoint());
         $this->clear();
--- a/src/Plugin/BufferedAdd/Event/AddDocument.php
+++ b/src/Plugin/BufferedAdd/Event/AddDocument.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Solarium\Plugin\BufferedAdd\Event;
 
 use Solarium\Core\Query\DocumentInterface;
--- a/src/Plugin/BufferedAdd/Event/Events.php
+++ b/src/Plugin/BufferedAdd/Event/Events.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Solarium\Plugin\BufferedAdd\Event;
 
 /**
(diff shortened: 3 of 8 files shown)


wickedOne:
which would mean you implicitly use the setter to set it to null?

thomascorthals:
At the least, I feel `@var` should reflect that `null` is legal value for the property on a fully initialised object.

> what's the use case for that?

Changing your mind after you've called the constructor? ;)

A very far-fetched use case might be a wizard where you construct the object and then call each setter based on user input for the next question, without bothering to check if the answer amounted to `null`. A little less far-fetched might be constructing the object once and using it in a loop where every iteration should be able to 'reset' the property if there is no meaningful value.

Not allowing `null` in the type hint could break BC with code that was written before type hinting was introduced in Solarium.

Bottom line for me is that it strikes me as odd that the constructor can do something the setter can't. Or to have the getter return something that can't be set.

I did notice there are only 3 setters in our codebase that allow `null` in the type hint. So I'm probably overthinking it. If there's any merit to my ramblings, they might be more suited in a code style discussion.

wickedOne:
yes, i agree with you on the var part and am mostly on your side when it comes to it being odd that the setters don't allow the return types.

the reason why i didn't change the setter typehints is because it's an event which normally is instantiated once and passed around mostly being immutable. in this specific case i think the buffer itself is the only thing which actually needs a setter

```
....
```

looking at the code once more, it doesn't really make sense
https://github.com/solariumphp/solarium/blob/9e073cc32c0144fc23215fe205cd0cb44e68b6ad/src/Plugin/BufferedAdd/BufferedAdd.php#L268
``addDocuments`` requires overwrite to be a boolean so the getter shouldn't be nullable nor should overwrite. apart from that it should call ``$event->getBuffer()`` rather than``$hits->buffer``

according to the docs [overwrite](https://lucene.apache.org/solr/guide/6_6/uploading-data-with-index-handlers.html#UploadingDatawithIndexHandlers-XMLFormattedIndexUpdates) defaults to true, i think in the bufferedAdd contructor argument it should default to true and we should be good.

what do you think?
i'll change my pr accordingly, thnx

thomascorthals:
I'm not sure about using Solr's default explicitly as our default. As it is now, if you don't set it explicitly, we don't pass it to Solr and whatever Solr's default behaviour is will happen. If we change that, we make explicit which behaviour we'll be requesting from Solr. If we choose the same default as Solr, it doesn't make a difference for our users _until Solr changes the default behaviour_. It has happened between major Solr releases, e.g. with `sow=false`.

It might depend on the user, or even on the user's use case for that behaviour, if he'd rather have the old or the new default for his existing code against an upgraded Solr. I don't think it's up to us to provide explicit defaults for every optional parameter. That would also mean that a user has to set Solr's default explicitly if it no longer matches our default choice. It will be confusing when not setting a parameter in Solarium results in the opposite behaviour of not setting it in a direct `curl` call to Solr.

wickedOne:
yes, point taken / makes sense
also forgot that the following is fine with php
```php
class A
{
    __construct(bool $b = null)
    {
    }
}

$b = new A(null);
```

as for the setters: you still prefer them to be nullable?

thomascorthals:
Since I'm being nitpicky, could you adjust the typehint in `PreFlush` as well? ;)
https://github.com/solariumphp/solarium/blob/8ec6d480ecb8b60cdb5c951f532ba051d2db5bd2/src/Plugin/BufferedAdd/Event/PreFlush.php#L18-L26

I would prefer the setters to be nullable to match the constructor functionality. It turns out this is the only place in our codebase where we have both nullable typehinted constructor arguments and setters for the same properties.

wickedOne:
k, changed typehints, annotations and setters for bufferedadd related event classes and declared strict types on them as well as we're being nitpicky :-)

PR #759, opened 2020-03-02
Integration test for Update query

thomascorthals:
Integration test for the current functionality of Update query, including raw XML commands.

### DIFF
--- a/tests/Integration/AbstractTechproductsTest.php
+++ b/tests/Integration/AbstractTechproductsTest.php
@@ -455,17 +455,6 @@ abstract class AbstractTechproductsTest extends TestCase
         $response = $this->client->update($update);
         $this->assertSame(0, $response->getStatus());
 
-        // add, rollback, commit
-        $update = $this->client->createUpdate();
-        $update->addDocument($doc1);
-        $this->client->update($update);
-        $update = $this->client->createUpdate();
-        $update->addRollback();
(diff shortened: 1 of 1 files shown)


thomascorthals:
I had to move the rollback test to a seperate method with `@group solr_no_cloud` because it isn't supported in SolrCloud mode ([SOLR-4895](https://issues.apache.org/jira/browse/SOLR-4895)). This means some double setup work.

thomascorthals:
> I had to move the rollback test to a seperate method with `@group solr_no_cloud` because it isn't supported in SolrCloud mode ([SOLR-4895](https://issues.apache.org/jira/browse/SOLR-4895)). This means some double setup work.

There's another way to check if we're testing a core or a collection without `@group`.

wickedOne:
you can probably do an api call retrieving that information, but why would you want to do that? having two different groups for solr cloud and solr standalone sounds like a valid workflow to me.

thomascorthals:
In this case, I don't like how I had to single out the rollback test in a different method with the `@group` approach. It felt like breaking up an integration into a smaller … unit. 😉 This seems like an elegant solution.
https://github.com/solariumphp/solarium/blob/a956603d270fc4edf241bd63b8d2c73f609d7bb6/tests/Integration/AbstractTechproductsTest.php#L459

PR #770, opened 2020-04-03
remove deprecated Endpoint timeout option

dmaicher:
Remove deprecated Endpoint timeout option + getter/setter

### DIFF
--- a/src/Core/Client/Adapter/TimeoutAwareTrait.php
+++ b/src/Core/Client/Adapter/TimeoutAwareTrait.php
@@ -8,9 +8,9 @@ namespace Solarium\Core\Client\Adapter;
 trait TimeoutAwareTrait
 {
     /**
-     * @var int|null
+     * @var int
      */
-    private $timeout;
+    private $timeout = TimeoutAwareInterface::DEFAULT_TIMEOUT;
 
     public function setTimeout(int $timeoutInSeconds): void
     {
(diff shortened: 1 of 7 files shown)


src/Core/Client/Adapter/TimeoutAwareTrait.php
```diff
@@ -8,9 +8,9 @@
 trait TimeoutAwareTrait
 {
     /**
-     * @var int|null
+     * @var int
      */
-    private $timeout;
+    private $timeout = TimeoutAwareInterface::DEFAULT_TIMEOUT;
 
     public function setTimeout(int $timeoutInSeconds): void
```
dmaicher, line 15: do we actually need this to be nullable? I think currently on the endpoint the timeout can be disabled (set to null)? or not?

src/Core/Client/Endpoint.php
```diff
@@ -188,37 +187,6 @@ public function getCore(): ?string
...
-    /**
-     * Set timeout option.
-     *
-     * @param int $timeout
-     *
-     * @return self Provides fluent interface
-     *
-     * @deprecated Endpoint::setTimeout is deprecated since Solarium 5.2 and will be removed in Solarium 6. Configure the timeout on the HTTP Client used by the Adapter instead.
-     */
-    public function setTimeout(int $timeout): self
```
dmaicher, line 200: ah no. The setter does not accept null anyway

Issue #774, opened 2020-04-06
Solarium 6: AdapterInterface and ConfigurableInterface


dmaicher:
Currently `AdapterInterface` extends `ConfigurableInterface` which is not a good idea imho.

The new `Psr18Adapter` for example does not make use of this at all.

To follow the "interface segregation principle" we should not force all adapters to implement `ConfigurableInterface`.

Remaining questions:
- do we rely anywhere in the code on master that all instances of `AdapterInterface` are also implementing `ConfigurableInterface`?
- as this would be a breaking change in 6.0: do we add deprecations (we discussed it [here](https://github.com/solariumphp/solarium/pull/754#issuecomment-593583513) also) on 5.x?

thomascorthals:
@wickedOne mentioned deprecating the `Configurable` god object in #752.

mkalkbrenner:
I managed to get search_api_solr to work with solarium 5.0 and 5.1. And with the current code base of the upcoming versions 5.2 and 6.0.

It is required to support these solarium versions in parallel because of the fact different solarium versions are used in Drupal 8 and 9.

As I mentioned in previous discussions we distinguish between different timeouts:
- search timeout
- index timeout
- optimize timeout
- finalization timeout

And these timeouts differ for different endpoints using the same adapter.

Finally I found an "easy" way to keep the existing functionality in a way compatible to all solarium 5 and 6 versions:

I set these four values as "options" on the endpoint objects and the adapter. The adapter holds the default values and the endpoints the concrete values that differ from the defaults.

This way I also use the ConfigurableInterface on the Psr18Adapter, too!

I know that the approach is kind of hacked. But other approaches would have caused much more code changes and incompatibilities for other modules that depend on search_api_solr.

I just want to mention this and ask for careful decisions here.

dmaicher:
@mkalkbrenner not sure I understand this point

> This way I also use the ConfigurableInterface on the Psr18Adapter, too!

So did you modify the `Psr18Adapter` so it uses its options?

mkalkbrenner:
Have a look at https://git.drupalcode.org/project/acquia_connector/-/blob/8.x-1.x/acquia_search/src/Plugin/SolrConnector/SearchApiSolrAcquiaConnector.php#L198 as example. this code need to be changed to use the Psr18Adapter.
(And there're other third party Solr service providers/hosters that have their own implementation of search_api_solr's Solr ConnectorInterface.)
[...]

PR #777, opened 2020-04-09
AdapterInterface does not extend ConfigurableInterface anymore

dmaicher:
Closes https://github.com/solariumphp/solarium/issues/774 

- For the `Http` and `Psr18Adapter` I see no reason to implement `ConfigurableInterface`
- `Curl` offers a `proxy` option so its still implementing `ConfigurableInterface`

I would not add any deprecations on 5.x for this because I actually see no way of having useful deprecations that can be removed before moving to 6.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Changed
 - More strict types and type hinting
+- `AdapterInterface` does not extend `ConfigurableInterface` anymore
+- `Http` Adapter does not implement `ConfigurableInterface` anymore
+- `Psr18Adapter` does not implement `ConfigurableInterface` anymore
 
 ### Removed
 - Zend2HttpAdapter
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Changed
 - More strict types and type hinting
+- `AdapterInterface` does not extend `ConfigurableInterface` anymore
+- `Http` Adapter does not implement `ConfigurableInterface` anymore
```
mkalkbrenner, line 15: The Http Adapter can handle timeouts:
```
'timeout' => $this->timeout ?? $endpoint->getTimeout(),
```
 So instead of removing the ConfigurableInterface, shouldn't we properly adopt it?

dmaicher: No the `Http` adapter does not support any options.

The line you mentioned is like this on master:
```php
'timeout' => $this->timeout,
```

and is just part of the context used for the http request? It has nothing to do with the `ConfigurableInterface`? I don't see any use of any option set on the adapter itself.

dmaicher: @mkalkbrenner or do you mean we should make the `timeout` configurable (via `ConfigurableInterface` and an option)? Not sure that is a good idea. We already have `TimeoutAwareInterface::setTimeout` for that.

mkalkbrenner: You're right.

PR #782, opened 2020-05-06
Input encoding can be something other than UTF-8

### LINKED ISSUE #367: Query parameter "ie"
Since version 4.5, Solr accepts a query parameter ie="<encoding>" which specify the encoding of the query text. If "ie" is supplied, SolrRequestParsers checks if the charset of the query text is the same of "ie" argument. If it is the same it decode it to UTF-8 (the standard encoding) otherwise throws a Bad Request exception. This is useful for PHP applications that uses "iso-8859-1" (very common) or other encoding types. Currently, Sollarium didn't implement it, right? Do you have plans to support it?

Very best regards,
Emanuel

thomascorthals:
Solr supports other input encodings than UTF-8. It could already be set in Solarium with a custom parameter for GET requests, but not for POST. I need to set it to import the `techproducts` testdata (suggested in #751) because not all sample XMLs are UTF-8 encoded. This PR closes #367 as well.

There is a parameter `ie` since Solr 4.5.0, but it isn't listed among the [Common Query Parameters](https://lucene.apache.org/solr/guide/8_5/common-query-parameters.html) in the Solr Ref Guide. The [changelog](https://github.com/apache/lucene-solr/blob/d4dbd0b9e75cb0bd5b0188a0f70070b6867fd94b/solr/CHANGES.txt#L12692-L12695) mentions it though.

```
* SOLR-5082: The encoding of URL-encoded query parameters can be changed with
  the "ie" (input encoding) parameter, e.g. "select?q=m%FCller&ie=ISO-8859-1".
  The default is UTF-8. To change the encoding of POSTed content, use the
  "Content-Type" HTTP header.  (Uwe Schindler, Shawn Heisey)
```

There's also this little tidbit from Solr's [source code](https://github.com/apache/lucene-solr/blob/d4dbd0b9e75cb0bd5b0188a0f70070b6867fd94b/solr/core/src/java/org/apache/solr/servlet/SolrRequestParsers.java#L340-L354).

```java
  public static final String INPUT_ENCODING_KEY = "ie";
  private static final byte[] INPUT_ENCODING_BYTES = INPUT_ENCODING_KEY.getBytes(CHARSET_US_ASCII);

// ...

            if (Arrays.equals(keyBytes, INPUT_ENCODING_BYTES)) {
              // we found a charset declaration in the raw bytes
              if (charsetDecoder != null) {
                throw new SolrException(ErrorCode.BAD_REQUEST,
                  supportCharsetParam ? (
                    "Query string invalid: duplicate '"+
                    INPUT_ENCODING_KEY + "' (input encoding) key."
                  ) : (
                    "Key '" + INPUT_ENCODING_KEY + "' (input encoding) cannot "+
                    "be used in POSTed application/x-www-form-urlencoded form data. "+
                    "To set the input encoding of POSTed form data, use the "+
                    "'Content-Type' header and provide a charset!"
                  )
                );
              }
```

From a (client library) user's standpoint, it makes sense to provide a single method `setInputEncoding()` and do the heavy lifting for them. In the case of `PostBigRequest`, they don't even know if they're sending a GET or a POST.

When an input encoding is set, I add a parameter `ie` to the query string. This works for GET and is safely ignored for POST. If the request is a POST, the same encoding is set in the `Content-Type` header.

You can not set the input encoding in POSTed form data. This will fail (for good reason, Solr needs to know the encoding before it starts parsing the data).
```bash
curl -X POST -H 'Content-type:application/x-www-form-urlencoded' -d 'ie=ISO-8859-1' http://localhost:8983/solr/techproducts/update
```
This doesn't seem to be an issue with Solarium. I don't believe there's a way to generate a request like this without a custom request builder.

For importing from XML files, the onus is on the user to set the matching input encoding for the file. If the XML declaration is correct, it can be determined automatically. But that would mean a lot of added complexity if you're importing multiple files that have different encodings, such as the `techproducts` testdata. Importing like that is a rather advanced feature that's mostly for our own benefit at this moment, so I don't think this is a dealbreaker.

I ran into a small issue with the generated XML for non-UTF-8 data because `htmlspecialchars()` was used to escape element content. It needs to know the encoding to work correctly and that information wasn't readily accessible at that point in the code. It's also technically wrong to use that function, it just happens to work because XML and HTML share a similar lineage from SGML. I've replaced it with a helper function that does the escaping.

I've added unit and integration tests with ISO-8859-1 as the non-UTF-8 charset. This is a deliberate choice. It's the only non-UTF charset (besides US-ASCII, but that'd be pointless for integration testing) in the [standard charsets](https://docs.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html#standard) that every Java platform is required to support. However, since `techproducts` contains an example in GB18030, I assume that it's also supported on all Solr instances. Once we import the testdata ourselves, this will be tested _de facto_.

The input encoding doesn't apply to JSON data. That can only be encoded using UTF-8, UTF-16 or UTF-32.

It should also be noted that it's **only** possible to change the input encoding for Solr. There's no corresponding `oe` parameter to change the output encoding.

### DIFF
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -474,6 +474,7 @@ class Helper
      */
     public function escapeXMLCharacterData(string $data): string
     {
+        // we don't use htmlspecialchars because it only supports a limited number of character sets
         return str_replace(['&', '<', '>'], ['&amp;', '&lt;', '&gt;'], $data);
     }
 
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -472,7 +472,7 @@ class Helper
      *
      * @return string
      */
-    public function escapePCDATAContent(string $data): string
+    public function escapeXMLCharacterData(string $data): string
     {
         return str_replace(['&', '<', '>'], ['&amp;', '&lt;', '&gt;'], $data);
     }
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -282,7 +282,6 @@ class Helper
     /**
      * Render a qparser plugin call.
      *
-     *
      * @param string $name
      * @param array  $params
      * @param bool   $dereferenced
@@ -462,8 +461,24 @@ class Helper
     }
 
     /**
-     * Render placeholders in a querystring.
+     * Escape text for use as parsed character data content in an XML element.
      *
+     * This escapes characters that can't appear as character data using their
+     * corresponding entity references. Per the definition of XML, "&" and "<"
+     * MUST be escaped when used in character data, ">" MUST be escaped in the
+     * string "]]>" and MAY be otherwise, so we escape it to be safe.
+     *
+     * @param string $data
+     *
+     * @return string
+     */
+    public function escapePCDATAContent(string $data): string
+    {
+        return str_replace(['&', '<', '>'], ['&amp;', '&lt;', '&gt;'], $data);
+    }
+
+    /**
+     * Render placeholders in a querystring.
      *
      * @param array $matches
      *
--- a/tests/Integration/AbstractTechproductsTest.php
+++ b/tests/Integration/AbstractTechproductsTest.php
@@ -779,7 +779,7 @@ abstract class AbstractTechproductsTest extends TestCase
         // output encoding: UTF-8 (always)
         $select->setQuery('cat:áéíóú');
         $result = $this->client->select($select);
-        $this->assertSame(1, $result->count());
+        $this->assertCount(1, $result);
         $this->assertSame([
             'id' => 'solarium-test-1',
             'name' => 'Sølåríùm Tëst 1',
@@ -791,7 +791,7 @@ abstract class AbstractTechproductsTest extends TestCase
         $select->setQuery('cat:'.utf8_decode('áéíóú'));
         $select->setInputEncoding('ISO-8859-1');
         $result = $this->client->select($select);
-        $this->assertSame(1, $result->count());
+        $this->assertCount(1, $result);
         $this->assertSame([
             'id' => 'solarium-test-1',
             'name' => 'Sølåríùm Tëst 1',
@@ -815,7 +815,7 @@ abstract class AbstractTechproductsTest extends TestCase
         $select->setQuery('cat:áéíóú');
         $select->setInputEncoding('UTF-8');
         $result = $this->client->select($select);
-        $this->assertSame(2, $result->count());
+        $this->assertCount(2, $result);
         $iterator = $result->getIterator();
         $this->assertSame([
             'id' => 'solarium-test-1',
@@ -834,7 +834,7 @@ abstract class AbstractTechproductsTest extends TestCase
         $update->addCommit(true, true);
         $this->client->update($update);
         $result = $this->client->select($select);
-        $this->assertSame(0, $result->count());
+        $this->assertCount(0, $result);
     }
 }
 
--- a/tests/Integration/AbstractTechproductsTest.php
+++ b/tests/Integration/AbstractTechproductsTest.php
@@ -516,15 +516,23 @@ abstract class AbstractTechproductsTest extends TestCase
         $result = $this->client->select($select);
         $this->assertCount(0, $result);
 
-        // add from files without and with Byte Order Mark and XML declaration
+        // add from UTF-8 encoded files without and with Byte Order Mark and XML declaration
         $update = $this->client->createUpdate();
         foreach (glob(__DIR__.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'testxml[1234]-add*.xml') as $file) {
             $update->addRawXmlFile($file);
         }
         $update->addCommit(true, true);
         $this->client->update($update);
+
+        // add from non-UTF-8 encoded file
+        $update = $this->client->createUpdate();
+        $update->setInputEncoding('ISO-8859-1');
+        $update->addRawXmlFile(__DIR__.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'testxml5-add-iso-8859-1.xml');
+        $update->addCommit(true, true);
+        $this->client->update($update);
+
         $result = $this->client->select($select);
-        $this->assertCount(4, $result);
+        $this->assertCount(5, $result);
         $iterator = $result->getIterator();
         $this->assertSame([
             'id' => 'solarium-test-1',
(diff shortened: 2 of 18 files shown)


src/Core/Query/Helper.php
```diff
@@ -462,8 +461,24 @@ public function filterControlCharacters(string $data): string
...
+     * MUST be escaped when used in character data, ">" MUST be escaped in the
+     * string "]]>" and MAY be otherwise, so we escape it to be safe.
      *
+     * @param string $data
+     *
+     * @return string
+     */
+    public function escapePCDATAContent(string $data): string
+    {
+        return str_replace(['&', '<', '>'], ['&amp;', '&lt;', '&gt;'], $data);
```
mkalkbrenner, line 477: I don't see an advantage here. According to the PHP manual, this function is exactly made for this:
`htmlspecialchars($value, ENT_NOQUOTES)`

src/Core/Query/Helper.php
```diff
@@ -462,8 +461,24 @@ public function filterControlCharacters(string $data): string
...
+     * This escapes characters that can't appear as character data using their
+     * corresponding entity references. Per the definition of XML, "&" and "<"
+     * MUST be escaped when used in character data, ">" MUST be escaped in the
+     * string "]]>" and MAY be otherwise, so we escape it to be safe.
      *
+     * @param string $data
+     *
+     * @return string
+     */
+    public function escapePCDATAContent(string $data): string
```
mkalkbrenner, line 475: Escaping as `CDATA` is the other option. But the this function doesn't do it.

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -743,6 +757,85 @@ public function testV2Api()
...
+        $doc->setField('price', 3.14);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        $this->client->update($update);
+
+        // input encoding: UTF-8 (default)
+        // output encoding: UTF-8 (always)
+        $select->setQuery('cat:áéíóú');
+        $result = $this->client->select($select);
+        $this->assertSame(1, $result->count());
```
mkalkbrenner, line 782: assertCount

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -743,6 +757,85 @@ public function testV2Api()
...
+            'name' => 'Sølåríùm Tëst 1',
+            'price' => 3.14,
+        ], $result->getIterator()->current()->getFields());
+
+        // input encoding: ISO-8859-1
+        // output encoding: UTF-8 (always)
+        $select->setQuery('cat:'.utf8_decode('áéíóú'));
+        $select->setInputEncoding('ISO-8859-1');
+        $result = $this->client->select($select);
+        $this->assertSame(1, $result->count());
```
mkalkbrenner, line 794: assertCount

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -743,6 +757,85 @@ public function testV2Api()
...
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        $this->client->update($update);
+
+        // input encoding: UTF-8 (explicit)
+        // output encoding: UTF-8 (always)
+        $select->setQuery('cat:áéíóú');
+        $select->setInputEncoding('UTF-8');
+        $result = $this->client->select($select);
+        $this->assertSame(2, $result->count());
```
mkalkbrenner, line 818: assertCount

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -743,6 +757,85 @@ public function testV2Api()
...
+            'name' => 'Sølåríùm Tëst 2',
+            'price' => 42.0,
+        ], $iterator->current()->getFields());
+
+        $update = $this->client->createUpdate();
+        $update->addDeleteQuery('cat:solarium-test');
+        $update->addCommit(true, true);
+        $this->client->update($update);
+        $result = $this->client->select($select);
+        $this->assertSame(0, $result->count());
```
mkalkbrenner, line 837: assertCount

src/Core/Query/Helper.php
thomascorthals: Actually, according to the PHP manual, this functions is:
`htmlspecialchars($value, ENT_XML1)`
I must admit I only found out just now.

My main problem with `htmlspecialchars` was that the encoding needs to be passed explicitly as a third parameter if it's something other than `ini_get("default_charset")`. It was used in places where I had no way to directly access the input encoding from the request.

Another possible issue is that's it's limited to a number of encodings. Not all standard charset for Java are supported. And I don't see GB18030 among the supported ones, which is used in the `techproducts` testdata.

However this turns out, having it in a helper function means we only have to change it in one place if we decide on a different implementation.

src/Core/Query/Helper.php
thomascorthals: Using `CDATA` sections isn't without pitfalls.

If the content of an element contains `]]>`, it has to be split up in multiple `CDATA` sections. Something like this will work:
```php
$value = '<![CDATA[' . str_replace(']]>', ']]]]><![CDATA[>', $value) . ']]>';
```

XML documents can potentially contain charachters that aren't available in the defined encoding by representing them as numeric references (`&#nnn;`). These characters cannot be represented in a `CDATA` section because it can only contain literal character data. Imported files might contain such references, so it's better to play it safe.

Transcoding during transport can also cause issues. I have no idea if this comes into play with Solr.

thomascorthals:
> We should just fix the small issues, add an entry to CHANGELOG and mention the parameter in the documentation.

Where do you propose to add it? I don't think other common query parameters (TZ, NOW …) are mentioned at the moment, or I would've added it there.

src/Core/Query/Helper.php
mkalkbrenner: In this case we should simply not name the function like this.

src/Core/Query/Helper.php
mkalkbrenner: Got it. In this case we extend the comments for this function and its purpose.

src/Core/Query/Helper.php
thomascorthals: Oh, I see the confusion now. I named it for [PCDATA](https://en.wikipedia.org/wiki/PCDATA), which does look a lot like [CDATA](https://en.wikipedia.org/wiki/CDATA).

mkalkbrenner:
You're right. We don't have a page in the documentation that explains common or global query parameters. Just a few pages contain a table of `Available options` at the top. I think it would be great to adapt that pattern and to create a global-request-parameters.md

src/Core/Query/Helper.php
thomascorthals: I've renamed the helper function to `escapeXMLCharacterData`. That should be less confusing.

src/Core/Query/Helper.php
thomascorthals: Should I add comments inside the function explaining why it doesn't use `htmlspecialchars`?

The PHPdoc already explains exactly what to expect from the function. From the caller's perspective, that's all that's needed. It's also exactly what the unit test is testing. Implementation details shouldn't matter on this level.

mkalkbrenner: We should add one sentence, why htmlspecialchars isn't used because not all character sets are supported.

thomascorthals: I've added the comment.

PR #785, opened 2020-05-19
Allow null return values in Solarium\Component\Result\Facet\Range (fixes #784)

### LINKED ISSUE #784: Fatal error: Uncaught TypeError in Solarium\Component\Result\Facet\Range
Hi,

If you call getAfter() on a facet result without the corresponding $facet->setOther('after') in the query you get a TypeError.
The same will happen with getBefore() and getBetween().
I couldn't find a way to know if the source query had the 'other' value set so I can't tell if it is 'safe' to call e.g. getAfter()

The constructor comments on Solarium\Component\Result\Facet\Range class explicitly state that the values could be null:
``` 
     * @param int|null        $before
     * @param int|null        $after
     * @param int|null        $between
```
I propose that the corresponding three functions should have their return type changed from int to ?int to allow null values, this is inline with the way that getPivot() is currently declared:
```
    public function getBefore(): ?int
    public function getAfter(): ?int
    public function getBetween(): ?int
```
Example:
if we have a set of 3 records:
simple_field:10
simple_field:20
simple_field:30

```
$query = $solr->createSelect();
$facetSet = $query->getFacetSet();
$facet = $facetSet->createFacetRange('simple');
$facet->setField('simple_field');
$facet->setStart(20);
$facet->setGap(10);
$facet->setEnd(30);
$facet->setOther('before');
$resultset = $solr->select($query);
```

```
$facet = $resultset->getFacetSet()->getFacet('simple');
$facet->getBefore(); # this works as we used setOther('before') on the query
$facet->getAfter(); # this causes the TypeError
PHP Fatal error:  Uncaught TypeError: Return value of Solarium\Component\Result\Facet\Range::getAfter() must be of the type integer, null returned in /vendor/solarium/solarium/src/Component/Result/Facet/Range.php:104
Stack trace:
#0 facet_range_test.php(40): Solarium\Component\Result\Facet\Range->getAfter()
#1 {main}
  thrown in /vendor/solarium/solarium/src/Component/Result/Facet/Range.php on line 104
```

I hope that makes sense? let me know if you need anymore information.
Many thanks,
Mark

ee-dev:
Solarium\Component\Result\Facet\Range constructor and get methods have different types.

```
     public function __construct(array $values, ?int $before, ?int $after, ?int $between, $start, $end, $gap, ?Pivot $pivot = null)
```
change the following getXXX methods to match:
```
    public function getBefore(): ?int
    public function getAfter(): ?int
    public function getBetween(): ?int
```

### DIFF
--- a/src/Component/Result/Facet/Range.php
+++ b/src/Component/Result/Facet/Range.php
@@ -94,9 +94,9 @@ class Range extends Field
      * Count of all records with field values lower then lower bound of the first range
      * Only available if the 'other' setting was used in the query facet.
      *
-     * @return int
+     * @return ?int
      */
-    public function getBefore(): int
+    public function getBefore(): ?int
     {
         return $this->before;
     }
@@ -107,9 +107,9 @@ class Range extends Field
      * Count of all records with field values greater then the upper bound of the last range
      * Only available if the 'other' setting was used in the query facet.
      *
-     * @return int
+     * @return ?int
      */
-    public function getAfter(): int
+    public function getAfter(): ?int
     {
         return $this->after;
     }
@@ -120,9 +120,9 @@ class Range extends Field
      * Count all records with field values between the start and end bounds of all ranges
      * Only available if the 'other' setting was used in the query facet.
      *
-     * @return int
+     * @return ?int
      */
-    public function getBetween(): int
+    public function getBetween(): ?int
     {
         return $this->between;
     }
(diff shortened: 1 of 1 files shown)


thomascorthals:
Good catch, by the way! If it can be set to `null`, it can be returned as `null`.

ee-dev:
> The proper way to document the nullable return type `?int` is `@return int|null`.

Duh! sorry, I even quoted from the constructor docs in the issue! Have fixed in my fork…should i do a new PR or can i update this one? (sorry this is my first public PR!)

thomascorthals:
Pushing the new commit to your own fork is fine. The PR was updated automatically with your changes. You can see this for yourself on the [Commits](https://github.com/solariumphp/solarium/pull/785/commits) tab of this PR.

mkalkbrenner:
Merged, thanks for the contribution :-)

PR #786, opened 2020-05-19
move to symfony contracts event dispatcher interface

### LINKED ISSUE #783: move to symfony contracts event dispatcher interface?
https://github.com/solariumphp/solarium/blob/beb674e36b461fc8824ce4f1e1bca2c8d6e5297c/src/Core/Client/Client.php#L62

are we ok with moving to the symfony contracts version of the event dispatcher interface?

wickedOne:
fixes #783


mkalkbrenner:
What is our goal?
Just switching the the interface or can we turn the dependency to the request dispatcher itself into an optional one?

PR #802, opened 2020-06-04
Examples and integration test for atomic updates

### LINKED ISSUE #499: Atomic Updates REMOVE modifier
Hi,

I realise this was discussed as issue #277 , but it appears something isn't right.

I've looked in the source code, and all that was added was the option to allow the constant MODIFIER_SET to be set to "remove", which doesn't actually appear to do anything anywhere else, other than not spit out an error.

I am attempting to remove an entry from a multi-value field. I've tried multiple ideas, but right now I'm using:

$document->removeField('mymultivalue', 'apple');
$document->setFieldModifier('mymultivalue', 'remove');
Of course 'apple' is indeed set for this particular multi-valued field.

But I'm not getting anywhere at all.

Also tried:

$document->setField('mymultivalue', 'apple');
$document->setFieldModifier('mymultivalue', 'remove');

$document->addField('mymultivalue', 'apple');
$document->setFieldModifier('mymultivalue', 'remove');

Maybe this wasn't implemented correctly to begin with, or maybe I am confused. Either way there doesn't seem to be much in the way of documentation for this feature after multiple searches. 

Can anyone please point me in the right direction?

Thanks for any responses!

thomascorthals:
Added a basic example of atomic updates (`set` and `inc` modifiers).

Updated the docs:
- Listed all available modifiers.
- Included the example.
- Mentioned that it depends on your schema.
- Updated the link to the latest Solr manual.

Integration test for all available modifiers.

This closes #499.

### DIFF
--- a/tests/Integration/AbstractTechproductsTest.php
+++ b/tests/Integration/AbstractTechproductsTest.php
@@ -664,6 +664,243 @@ abstract class AbstractTechproductsTest extends TestCase
         $this->assertSame(0, json_decode($response->getBody())->responseHeader->status);
     }
 
+    public function testModifiers()
+    {
+        $select = self::$client->createSelect();
+        $select->setQuery('id:solarium-test');
+        $select->addSort('id', $select::SORT_ASC);
+        $select->setFields('id,name,cat,price');
+        $update = self::$client->createUpdate();
+
+        $doc = $update->createDocument();
+        $doc->setField('id', 'solarium-test');
+        $doc->setField('name', 'Solarium Test');
+        $doc->setField('cat', 'solarium-test');
+        $doc->setField('price', 17.01);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
+            'id' => 'solarium-test',
+            'name' => 'Solarium Test',
+            'cat' => [
+                'solarium-test',
+            ],
+            'price' => 17.01,
+        ], $result->getIterator()->current()->getFields());
+
+        // set
+        $doc = $update->createDocument();
+        $doc->setKey('id', 'solarium-test');
+        $doc->setField('cat', 'modifier-set');
+        $doc->setFieldModifier('cat', $doc::MODIFIER_SET);
+        $doc->setField('price', 42.0);
+        $doc->setFieldModifier('price', $doc::MODIFIER_SET);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
+            'id' => 'solarium-test',
+            'name' => 'Solarium Test',
+            'cat' => [
+                'modifier-set',
+            ],
+            'price' => 42.0,
+        ], $result->getIterator()->current()->getFields());
+
+        // add & inc
+        $doc = $update->createDocument();
+        $doc->setKey('id', 'solarium-test');
+        $doc->setField('cat', 'modifier-add');
+        $doc->setFieldModifier('cat', $doc::MODIFIER_ADD);
+        $doc->setField('price', 5);
+        $doc->setFieldModifier('price', $doc::MODIFIER_INC);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
+            'id' => 'solarium-test',
+            'name' => 'Solarium Test',
+            'cat' => [
+                'modifier-set',
+                'modifier-add',
+            ],
+            'price' => 47.0,
+        ], $result->getIterator()->current()->getFields());
+
+        // add vs. add-distinct
+        $doc = $update->createDocument();
+        $doc->setKey('id', 'solarium-test');
+        $doc->setField('cat', ['modifier-add', 'modifier-add-another']);
+        $doc->setFieldModifier('cat', $doc::MODIFIER_ADD);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
+            'id' => 'solarium-test',
+            'name' => 'Solarium Test',
+            'cat' => [
+                'modifier-set',
+                'modifier-add',
+                'modifier-add',
+                'modifier-add-another',
+            ],
+            'price' => 47.0,
+        ], $result->getIterator()->current()->getFields());
+        $doc = $update->createDocument();
+        $doc->setKey('id', 'solarium-test');
+        $doc->setField('cat', ['modifier-add', 'modifier-add-another', 'modifier-add-distinct']);
+        $doc->setFieldModifier('cat', $doc::MODIFIER_ADD_DISTINCT);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
+            'id' => 'solarium-test',
+            'name' => 'Solarium Test',
+            'cat' => [
+                'modifier-set',
+                'modifier-add',
+                'modifier-add',
+                'modifier-add-another',
+                'modifier-add-distinct',
+            ],
+            'price' => 47.0,
+        ], $result->getIterator()->current()->getFields());
+
+        // remove & negative inc
+        $doc = $update->createDocument();
+        $doc->setKey('id', 'solarium-test');
+        $doc->setField('cat', 'modifier-set');
+        $doc->setFieldModifier('cat', $doc::MODIFIER_REMOVE);
+        $doc->setField('price', -5);
+        $doc->setFieldModifier('price', $doc::MODIFIER_INC);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
+            'id' => 'solarium-test',
+            'name' => 'Solarium Test',
+            'cat' => [
+                'modifier-add',
+                'modifier-add',
+                'modifier-add-another',
+                'modifier-add-distinct',
+            ],
+            'price' => 42.0,
+        ], $result->getIterator()->current()->getFields());
+
+        // remove multiple
+        $doc = $update->createDocument();
+        $doc->setKey('id', 'solarium-test');
+        $doc->setField('cat', ['modifier-add', 'modifier-add-another']);
+        $doc->setFieldModifier('cat', $doc::MODIFIER_REMOVE);
+        $update->addDocument($doc);
+        $update->addCommit(true, true);
+        self::$client->update($update);
+        $result = self::$client->select($select);
+        $this->assertCount(1, $result);
+        $this->assertSame([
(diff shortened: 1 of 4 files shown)


thomascorthals:
It looks like `add-distinct` doesn't work in the cloud tests. Any idea why? I don't have a local cloud setup to test against, and I didn't think it would matter.

mkalkbrenner:
The tests fail on Solr 7 in cloud mode:
https://github.com/solariumphp/solarium/pull/802/checks?check_run_id=740044161

thomascorthals:
@mkalkbrenner It looks like `add-distinct` still adds values that were already in the list in the cloud tests. It correctly ignores them in the server tests. I'm only able to run the server tests on my local machine so I can't really tell what's going wrong.

mkalkbrenner:
@thomascorthals there're docker-compose.yml files in the repository

mkalkbrenner:
https://github.com/solariumphp/solarium/blob/master/tests/Integration/Fixtures/docker/solr7_cloud/docker-compose.yml

thomascorthals:
It fails in Solr 7 cloud but passes in Solr 8 cloud on my local installation as well.

The implementation of `doAddDistinct` hasn't changed between versions: https://github.com/apache/lucene-solr/blame/0c4d8fb116f7dec83a04d27df550640dcec4eb8d/solr/core/src/java/org/apache/solr/update/processor/AtomicUpdateDocumentMerger.java#L445

mkalkbrenner:
I faced such situations multiple times over the last years and that is why I encourage these integration tests with real Solr servers.
Often this is caused by something else in Solr, like new or changed field types, something in solrconf.xml, ...
In rare cases I had to implement swicth statements in Search API Solr.
If I find some time I'll take a look. Did you allow code editing in this PR?

thomascorthals:
> Did you allow code editing in this PR?

@mkalkbrenner I did.

thomascorthals:
```shell
./vendor/bin/phpunit --no-coverage --repeat 10 --filter testModifiers tests/Integration/SolrCloud/Psr18Test.php
```

With Solr 8 cloud, they all pass. With Solr 7 cloud, some repeats fail and some pass. The first one fails most of the time, but not always. I can't discern a pattern as to which repeats pass, it's different every time I run the command.

Can anyone confirm this?

mkalkbrenner:
± ./vendor/bin/phpunit --no-coverage --repeat 10 --filter testModifiers tests/Integration/SolrCloud/Psr18Test.php
PHPUnit 8.5.5 by Sebastian Bergmann and contributors.

..........                                                        10 / 10 (100%)

Time: 35 seconds, Memory: 8.00 MB

OK (10 tests, 230 assertions)

mkalkbrenner:
So, locally the tests always pass on solr8_cloud

mkalkbrenner:
On solr7_cloud I get random test failures:

± ./vendor/bin/phpunit --no-coverage --repeat 10 --filter testModifiers tests/Integration/SolrCloud/Psr18Test.php
PHPUnit 8.5.5 by Sebastian Bergmann and contributors.

...F.F.FFF                                                        10 / 10 (100%)

Time: 38.11 seconds, Memory: 10.00 MB

There were 5 failures:

mkalkbrenner:
The solr7_cloud container seems to have serious issues after the tests!
Be aware that the console output is just the one of the second node running on port 8987 which is not exposed. To see the log of the first node on 8983 you need to log into the container.

thomascorthals:
@mkalkbrenner That's the same behaviour I'm seeing, across the three adapters.

mkalkbrenner:
One node seems to "die". That explains the 50% result as you have a 50% chance to hit the one that is alive. I assume the second one dies when the command gets replicated for the first time.

mkalkbrenner:
you can follow both logs by running these commands in parallel:

`docker exec -it solr7_cloud_solr7_1 bash -c "tail -f /opt/solr/server/logs/solr.log"`
`docker exec -it solr7_cloud_solr7_1 bash -c "tail -f /opt/solr/server/logs/solr.log.1"`

mkalkbrenner:
I repeated the tests now multiple times after I removed the existing containers. Now the nodes "survive". So I assume that my first first observation was unrelated.
Even if 50% of the tests fail I can monitor both nodes and access the UI.

So it look like a kind of replication drift or not flushed caches issue, meaning you write to one one node and immediately read from the other in 50% of the cases.

mkalkbrenner:
https://lucene.apache.org/solr/guide/7_0/shards-and-indexing-data-in-solrcloud.html#ignoring-commits-from-client-applications-in-solrcloud

I pretty sure it is a commit / replication issue. the default auto commit time in Solr 7's techproducts config set is 15s.

But the tests don't pass even if I add a 20s sleep after each update. But openSearcher is also set to "false".

So I suggest to try to adjust the config in AbstractCloudTest::createTechproducts() using the Config API and to tweak the autoCommit settings.
See https://lucene.apache.org/solr/guide/7_7/config-api.html#commands-to-modify-the-config

thomascorthals:
I ran it on a stock Solr 7.7.2 on Windows 10 instead of the Docker image. So it's a problem with the default config, not something we did wrong with our setup.

A replication issue had crossed my mind. Although I have no idea why it would only happen with `add-distinct`.

thomascorthals:
I had already disabled autoCommit and autoSoftCommit in `testUpdate`. I've moved that to `setUpBeforeClass` to disable it for all tests. Didn't bother to reset to the defaults any longer because we use throwaway cores/collections now.

Unfortunately, the problem still remains.

tests/Integration/AbstractTechproductsTest.php
```diff
@@ -55,6 +55,22 @@ public static function setUpBeforeClass(): void
         $ping = self::$client->createPing();
         self::$client->ping($ping);
 
+        // disable automatic commits for update tests
+        $request = new Request();
```
mkalkbrenner, line 59:
```php
$query = self::$client->createApi([
    'version' => Request::API_V2,
    'handler' => 'config',
    'rawdata' => json_encode([
            'set-property' => [
                'updateHandler.autoCommit.maxDocs' => -1,
                'updateHandler.autoCommit.maxTime' => -1,
                'updateHandler.autoSoftCommit.maxDocs' => -1,
                'updateHandler.autoSoftCommit.maxTime' => -1,
            ],
        ]),
]);
$response = self::$client->execute($query);
```

thomascorthals: I had to use `API_V1`. V2 only has a config endpoint for collections, not for cores.

I've set `openSearcher` explicitly as well and included a check that reads the properties back and ensures they are what we want them to be.

thomascorthals:
We might have discovered a bug in Solr 7. It looks like an `add-distinct` is replicated as an `add` to the other node.

We start with:
```
'cat' => [
    0 => 'modifier-set',
    1 => 'modifier-add',
    2 => 'modifier-add',
    3 => 'modifier-add-another',
],
```

We try and add `['modifier-add', 'modifier-add-another', 'modifier-add-distinct']` with `MODIFIER_ADD_DISTINCT` which should ignore `'modifier-add', 'modifier-add-another'` because these values already exist in the field. We expect, and sometimes get:
```
'cat' => [
    0 => 'modifier-set',
    1 => 'modifier-add',
    2 => 'modifier-add',
    3 => 'modifier-add-another',
    4 => 'modifier-add-distinct',
],
```

When the test fails, we actually end up with:
```
'cat' => [
    0 => 'modifier-set',
    1 => 'modifier-add',
    2 => 'modifier-add',
    3 => 'modifier-add-another',
    4 => 'modifier-add',
    5 => 'modifier-add-another',
    6 => 'modifier-add-distinct',
],
```

The only way I can see that happening is _if the `add-distinct` is replicated as a non-distinct `add` to the other node_.

thomascorthals:
I tried to reproduce it with `curl` requests from the shell to post to the Solr mailing list. And now I'm wondering how we manage to have the test pass at all. With my test scenario, both nodes didn't add distinctly.

**Solr 7.7.3, single core, techproducts configset:**

```shell
$ curl -X POST -H 'Content-Type: text/json' 'http://localhost:8983/solr/techproducts/update?commit=true' --data-binary '[{"id":123,cat:["a","b","c"]}]'
{
  "responseHeader":{
    "status":0,
    "QTime":136}}

$ curl 'http://localhost:8983/solr/techproducts/select?q=id%3A123&omitHeader=true'
{
  "response":{"numFound":1,"start":0,"docs":[
      {
        "id":"123",
        "cat":["a",
          "b",
          "c"],
        "_version_":1668928736812072960}]
  }}

$ curl -X POST -H 'Content-Type: text/json' 'http://localhost:8983/solr/techproducts/update?commit=true' --data-binary '[{"id":123,cat:{"add-distinct":["a","b","d"]}}]'
{
  "responseHeader":{
    "status":0,
    "QTime":70}}

$ curl 'http://localhost:8983/solr/techproducts/select?q=id%3A123&omitHeader=true'
{
  "response":{"numFound":1,"start":0,"docs":[
      {
        "id":"123",
        "cat":["a",
          "b",
          "c",
          "d"],
        "_version_":1668928797140844544}]
  }}
```

**Solr 7.7.3, cluster with 2 nodes, collection with shard count 1 and replicationFactor 2, techproducts configset**

```shell
$ curl -X POST -H 'Content-Type: text/json' 'http://localhost:8983/solr/techproducts/update?commit=true' --data-binary '[{"id":123,cat:["a","b","c"]}]'
{
  "responseHeader":{
    "rf":2,
    "status":0,
    "QTime":75}}

$ curl 'http://localhost:8983/solr/techproducts/select?q=id%3A123&omitHeader=true'
{
  "response":{"numFound":1,"start":0,"docs":[
      {
        "id":"123",
        "cat":["a",
          "b",
          "c"],
        "_version_":1668919529371074560}]
  }}

$ curl 'http://localhost:8987/solr/techproducts/select?q=id%3A123&omitHeader=true'
{
  "response":{"numFound":1,"start":0,"docs":[
      {
        "id":"123",
        "cat":["a",
          "b",
          "c"],
        "_version_":1668919529371074560}]
  }}

$ curl -X POST -H 'Content-Type: text/json' 'http://localhost:8983/solr/techproducts/update?commit=true' --data-binary '[{"id":123,cat:{"add-distinct":["a","b","d"]}}]'
{
  "responseHeader":{
    "rf":2,
    "status":0,
    "QTime":81}}

$ curl 'http://localhost:8983/solr/techproducts/select?q=id%3A123&omitHeader=true'
{
  "response":{"numFound":1,"start":0,"docs":[
      {
        "id":"123",
        "cat":["a",
          "b",
          "c",
          "a",
          "b",
          "d"],
        "_version_":1668919799351083008}]
  }}

$ curl 'http://localhost:8987/solr/techproducts/select?q=id%3A123&omitHeader=true'
{
  "response":{"numFound":1,"start":0,"docs":[
      {
        "id":"123",
        "cat":["a",
          "b",
          "c",
          "a",
          "b",
          "d"],
        "_version_":1668919799351083008}]
  }}
```

It's also worth noting that this only happens with `cat:{"add-distinct":["a","b","d"]}`. The behaviour for `cat:{"add-distinct":"a"}` is correct in cloud mode.

thomascorthals:
I've posted my minimal testcase to the [solr-user mailing list](http://mail-archives.apache.org/mod_mbox/lucene-solr-user/202006.mbox/%3CCAB9nurqW6RpxG1VZ1vNCuDbTjc9_1dTtLp7zt7vLXoJVAjAFLA%40mail.gmail.com%3E).

thomascorthals:
Added a check for `add-distinct` of a single value. `add` and `remove` were already checked separately for single and multiple values.

That way we have at least one check for `add-distinct` that works on Solr 7 cloud. ;)

mkalkbrenner:
I committed an intermediate fix for the phpstan issue. So the tests are running again now.
Unfortunately they still fail for Solr 7 Cloud.

PR #805, opened 2020-06-08
add conflict with phpstan 0.12.26 and 0.12.27

dmaicher:
PHPstan switched to AST based static analysis and there seem to be some issues with that.

The fail is actually wrong:

```
 ------ ---------------------------------------------------------------------------------- 
  Line   src/Component/Facet/JsonRange.php                                                 
 ------ ---------------------------------------------------------------------------------- 
  38     Call to an undefined method Solarium\Component\Facet\JsonRange::jsonFacetInit().  
 ------ ---------------------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 1 error   
```  

Reported the bug: https://github.com/phpstan/phpstan/issues/3415

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,5 +43,8 @@
         "psr-4": {
             "Solarium\\Tests\\": "tests/"
         }
+    },
+    "conflict": {
+        "phpstan/phpstan": "0.12.26 || 0.12.27"
     }
 }


thomascorthals:
Thanks!

My PR that fails because of a possible Solr bug now also fails because of a phpstan bug. ;)

composer.json
```diff
@@ -43,5 +43,8 @@
         "psr-4": {
             "Solarium\\Tests\\": "tests/"
         }
+    },
+    "conflict": {
```
dmaicher, line 47: ah actually this is not good I think. It will prevent those versions to be installed for anyone **using** solarium. We only need this for "dev" here...

dmaicher:
For now I guess we can only pin the exact version `0.12.25` until a new fix is released? :cry:

PR #772, opened 2020-04-03
remove deprecated methods for local parameters

wickedOne:
deprecated convenience methods for local parameters in preparation
for solarium 6.0

fixes #771

PR #811, opened 2020-06-15
fixes #810 by adding numBucket support to JSON facet results

### LINKED ISSUE #810: Feature Request: include numBuckets in JSON facet results
JSON terms facets include the ability to get a total number of buckets, irrespective of the number requested by 'limit', by including 'numBuckets':true.
Solarium includes this in the request and passes on to Solr but does nothing with the response.
If you set the facet terms limit to 0 then, currently, you don't get a Buckets object in the response.
I propose that you if you ask for numBuckets then you should get a Buckets object with an empty array of Bucket alongside numBuckets.

`Solarium\Component\Result\Facet\Buckets`
- Store numBuckets `protected $numBuckets` int|null
- Add numBuckets to the constructor `__construct(array $buckets, $numBuckets = null)`
- Add method `getNumBuckets() ?int`

`Solarium\Component\ResponseParser\FacetSet`
- Check for numBuckets in response i.e. `isset($values['numBuckets'])`
- If we have some Buckets return `new Buckets($buckets,$numBuckets);`
- If there are no buckets BUT we do have numBuckets then return Buckets(array(),$ numBuckets);

I have also included a simple example using a json term facet in the PR

ee-dev:
added numBucket to Buckets class and changed FacetSet to use numBucket response.
created example to use jsonTerms


ee-dev:
Apologies for delay — my phone line and therefore broadband are down. 
Will do my best to write the tests — I'm also working on examples for json query and json range facets but looks like these might require some more changes to parseJsonFacetSet so will post separate issue.

examples/2.1.5.1.7-facet-json-terms.php
```diff
@@ -0,0 +1,48 @@
...
+$client = new Solarium\Client($adapter, $eventDispatcher, $config);
+
+// get a select query instance
+$query = $client->createSelect();
+
+// get the facetset component
+$facetSet = $query->getFacetSet();
+
+// create a json terms instance and set options
+$categoriesTerms = new Solarium\Component\Facet\JsonTerms(['key' => 'categories', 'field' => 'cat', 'limit'=>4,'numBuckets'=>true]);
```
mkalkbrenner, line 16: I that this doesn't work anymore in solarium 6 since the commit of #772.
You need to name 'key' as 'local_key'.

Luckily I recently integrated the execution of the examples in the test pipeline ;-)

ee-dev: Thanks, am using a composer version of 5 in my project, will try switching to 6.

mkalkbrenner:
So we're almost done. But we need to
* adjust CHANGELOG.md
* extend an existing result parser unit test

src/Component/Result/Facet/Buckets.php
```diff
@@ -18,14 +18,22 @@ class Buckets implements FacetResultInterface, \IteratorAggregate, \Countable
...
+     */
+    protected $numBuckets;
+
     /**
      * Constructor.
      *
      * @param Bucket[] $buckets
      */
-    public function __construct(array $buckets)
+    public function __construct(array $buckets, $numBuckets = null)
```
thomascorthals, line 33: We're down to nitpicking (which is always a good sign!): `$numBuckets` is missing an `@param` and a typehint.

ee-dev: Thank you for your patience guiding me through the process. I think we're nearly there! Will make sure to re-read this PR before future submissions as it now lists all the things i forgot to do ;-)

Issue #747, opened 2020-02-04
Coding style adjustments?


dmaicher:
While working on the first PR I noticed a few things:
- void return types not consistent throughout the codebase
    - https://github.com/solariumphp/solarium/blob/master/tests/QueryType/Select/Query/FilterQueryTest.php#L17
    - https://github.com/solariumphp/solarium/blob/master/tests/Core/Query/LocalParameters/LocalParametersTest.php#L74
- lots of unnecessary phpdoc for stuff that is expressed with type-hints already
    - https://github.com/solariumphp/solarium/blob/master/src/Core/Query/LocalParameters/LocalParameters.php#L36
- some `@throws` phpdocs that seem mostly useless:
    - https://github.com/solariumphp/solarium/blob/master/tests/Core/Query/LocalParameters/LocalParametersTest.php#L22

Are you open for adjusting this? Mostly this can be automated anyway :blush: Whats the opinion of the maintainers?

wickedOne:
personally, on the phpdoc and @throws annotations, i prefer having them.

not only does it the ease of code review, but typehinting for example
```php
 // ...
 *  @return \Solarium\Component\Result\Facet\Pivot\Pivot[]
 // ...
```
instead of having
```php
// ...
public function getPivots(): array
// ...
```
has quite some benefits when using for instance phpstorm as your ide.

i'm not sure what you mean with a useless throw annotation. if an exception is thrown in a method i'd prefer to have that annotated. if you mean that annotations and typehinting is rather useless in unit tests, i could agree with you as there's no ide benefit, but on the other hand it serves the purpose of consistency in the library so i'm somewhat on the fence on that one

dmaicher:
The phpdoc you mention above makes perfect sense for me :+1: I was talking more about 

```php
@return string
```

or similar cases that bring zero additional value as the information is in typehints like 

```php
function foo(): string 
```

already. So I mean to remove duplicated phpdocs only.

Regarding the exceptions: yes mainly I find them useless in the tests :smile: No one will ever call phpunit test methods and care if it throws something or not. Its adds a lot of visual depth and is just more code to read. But maybe thats just my opinion.

wickedOne:
that would result in some methods with annotations and some without, which in my opion makes the codebase messier than anything else (again referring to code review and reading code  in general).

nevertheless, codestyles are there to get rid of opinions :-)
> Solarium follows the Symfony2 code standards: http://symfony.com/doc/current/contributing/code/standards.html

[...]

PR #817, opened 2020-06-22
Exceptions: consistency, unit tests, docs, examples

thomascorthals:
As discussed in #747, there were some inconsistencies in the way Solarium threw exceptions.

## TL;DR

- I've added `LogicExceptionInterface` and `RuntimeExceptionInterface`
- and made sure all thrown exceptions are "ours"
- while maintaining full backward compatibility,
- added examples and docs on catching them,
- and provided unit tests to prove our code and docs line up.

## What I didn't do

I didn't change the inheritance tree of the exceptions.
- Building our own tree under `Solarium\Exception` would break BC for users who are e.g. catching `Solarium\Exception\OutOfBoundsException` as a base `OutOfBoundsException`.
- Having `Solarium\Exception\StreamException` extend `Solarium\Exception\RuntimeException` would break BC for users who are catching it as a base `UnexpectedValueException`.
- Having `Solarium\Exception\HttpException` extend `Solarium\Exception\RuntimeException` wouldn't break BC, but doesn't seem terribly consistent at this point.

## What I did do

I've added `LogicExceptionInterface` and `RuntimeExceptionInterface` marker interfaces that extend `ExceptionInterface`. This adds a level of granularity in distinguishing exceptions while maintaining BC.

I've replaced the handful of base exceptions that were thrown with "proper" Solarium exceptions. They extend their base counterpart so this doesn't break BC.

I've added `use` statements for all exceptions in `src` and `tests` instead of scattering fully qualified names throughout the code.

I've added an example that demonstrates how you can distinguish Solarium exceptions from those thrown by other libraries. I've also written short documentation, emphasising that you must at least catch an `HttpException` for every request.

Examples that used `catch (Solarium\Exception $e)` have been updated to `catch (Exception $e)`. They ran fine as long as there is no exception, but would result in a fatal error when there is because they didn't actually match the exception we are trying to catch.

Since I've "formalised" the way to handle our exceptions in the docs, I've written unit tests that prove the inheritance model is what we describe. This safeguards us against introducing backward incompatible changes to the inheritance tree.

Test coverage for `HttpException` and `StreamException` has been rounded out to cover all method signatures.


mkalkbrenner:
I thought twice about this patch.
Since it maintains BC and increases the quality I'm of the opinion that we should include it in the final solarium 6.0.0 release even if we already reached rc1.

Issue #819, opened 2020-06-24
Getting named facet from a JsonQuery using getFacet('local_key_value') throws TypeError


ee-dev:
JsonQuery result is parsed into a FacetSet but FacetSet is not a 'FacetResultInterface'

Build query:
```
$facetSet = $query->getFacetSet();
$inStock = new Solarium\Component\Facet\JsonQuery(['local_key' => 'stock_query', 'query' => 'inStock:true']);
$facetSet->addFacet($inStock);
```
Send the query and parse the result:

```
$resultset = $client->select($query);
$resultset->getFacetSet()-getFacet('stock_query');
```
Throws Exception:

```
TypeError: Return value of Solarium\Component\Result\FacetSet::getFacet() must implement interface Solarium\Component\Result\Facet\FacetResultInterface or be null, instance of Solarium\Component\Result\FacetSet
```

Proposed fix would be to add FacetResultInterface to the FacetSet class
```
class FacetSet implements \IteratorAggregate, \Countable , FacetResultInterface
```
The current workaround is to iterate until you match the facet you want:
```
$facets->$resultset->getFacetSet();
foreach($facets as $key=>$facet)
{
	if($key == 'stock_query') {
		$stock_query = $facet;
		break;
		}
}

$stock_query_value = $stock_query->getFacet('count')->getValue();
```
Note that the workaround would be unaffected if we add FacetResultInterface to the FacetSet class interface list.

PR #822, opened 2020-06-26
Fix solariumphp/solarium#819 Getting named facet from a JsonQuery using getFacet('local_key_value') throws TypeError  

ee-dev:
Add FacetResultInterface to allow Solarium\Component\Result\FacetSet to access to a JsonQuery result by it's key in FacetSet::getFacet()
Includes JsonQuery example.


examples/index.html
```diff
@@ -56,6 +56,7 @@ <h2>Examples</h2>
                             <li><a href="2.1.5.1.5-facet-pivot.php">2.1.5.1.5 Facet pivot</a></li>
                             <li><a href="2.1.5.1.6-facet-interval.php">2.1.5.1.6 Facet interval</a></li>
                             <li><a href="2.1.5.1.7-facet-json-terms.php">2.1.5.1.7 Facet json terms</a></li>
+                            <li><a href="2.1.5.1.8-facet-json-query">2.1.5.1.8 Facet json query</a></li>
```
thomascorthals, line 59: Can you add `.php` to the link? It depends on the webserver config if it'll work without the extension.

```suggestion
                            <li><a href="2.1.5.1.8-facet-json-query.php">2.1.5.1.8 Facet json query</a></li>
```

examples/2.1.5.1.8-facet-json-query.php
```diff
@@ -0,0 +1,45 @@
...
+// this executes the query and returns the result
+$resultset = $client->select($query);
+
+// display the total number of documents found by solr
+echo 'NumFound: '.$resultset->getNumFound();
+
+// JsonQuery returns a FacetSet, not Buckets like a jsonTerms facet
+$stock_query = $resultset->getFacetSet()->getFacet('stock_query');
+
+// the facet count is returned as a 'count' sub-facet
```
mkalkbrenner, line 30: In search_api_solr we use this comment:
    
```
// The implicit "count" aggregation over all results matching the query
// exists only any JSONFacet set.
```

I think we should comment like this:
```
// A JSONFacet triggers an implicit "count" aggregation over all results which is accessible
// just like a Facet within the FacetSet.
```

ee-dev: Ah, except I'm not sure 'all results' is quite the right phrase as this is the count for the JsonQuery alone, not all results?
Perhaps:
// A JsonQuery triggers an implicit "count" aggregation for the result of the query which is accessible just like a Facet within the FacetSet.

mkalkbrenner: sounds good :-)

PR #824, opened 2020-06-26
Update requirements for Solarium 6.0 in "Getting started" docs

thomascorthals:


docs/getting-started.md
```diff
@@ -9,9 +9,9 @@ Installation
 
 ### Requirements
 
-For installing Solarium a minimal PHP version 7.1 is required.
+For installing Solarium a minimal PHP version 7.2 is required.
 
-There is no Solr version requirement. But it's highly recommended that you use at least 6.6.6.
+There is no Solr version requirement. But it's highly recommended that you use at least 7.7. All older version are EOL.
```
wickedOne, line 14: perhaps change it to something more generic?

> There is no Solr version requirement though solarium cannot guarantee compatibility with Solr versions which are EOL. Therefor it's highly recommended that you use the latest Solr version available.

mkalkbrenner: I think we should be specific here.

Issue #823, opened 2020-06-26
Results of before/after/between are not parsed from JsonRange response


ee-dev:
When using a JsonRange there is no way to get to the counts for 'before', 'after' or 'between'.

Build query:
```
$facetSet = $query->getFacetSet();
$priceranges = new \Solarium\Component\Facet\JsonRange(['local_key' => 'priceranges', 'field' => 'price','start'=>1 ,'end'=>300,'gap'=>100,'other'=>'all']);
$facetSet->addFacet($priceranges);
```
Send the query and parse the result:
```
$resultset = $client->select($query);
$facet = $resultset->getFacetSet()->getFacet('priceranges');
```
$facet is an instance of Solarium\Component\Result\Facet\Buckets with an array of 3 Bucket objects:
```
Solarium\Component\Result\Facet\Buckets Object
(
    [buckets:protected] => Array
        (
            [0] => Solarium\Component\Result\Facet\Bucket Object
                (
                    [value:protected] => 1
                    [count:protected] => 4
                )

            [1] => Solarium\Component\Result\Facet\Bucket Object
                (
                    [value:protected] => 101
                    [count:protected] => 2
                )

            [2] => Solarium\Component\Result\Facet\Bucket Object
                (
                    [value:protected] => 201
                    [count:protected] => 1
                )
        )

    [numBuckets:protected] => 
)
```
However the Solr Result is:
```
"facets":{
    "count":32,
    "priceranges":{
      "buckets":[ {"val":1.0, "count":4}, {"val":101.0, "count":2}, {"val":201.0,"count":1} ],
      "before":{"count":3},
      "after":{"count":6},
      "between":{"count":7}
    }
}
```

While I can probably make the needed changes to parseJsonFacetSet I don't know how best to return the results. I see two primary options:

### OPTION 1
 The simplest would be to create a subclass of Buckets e.g. `Solarium\Component\Result\Facet\JsonRange`
It would need three additional properties 'before', 'after' and 'between' alongside their respective getXXX methods.
This is similar to how non-json Ranges are handled.
The problem with this is a new object type is returned which _might_ break existing code.
```
Solarium\Component\Result\Facet\JsonRange Object
(
    [before:protected] = > 3
    [after:protected] = > 6
    [between:protected] = > 7
    [buckets:protected] => Array
        (
            [0] => Solarium\Component\Result\Facet\Bucket Object
                (
                    [value:protected] => 1
                    [count:protected] => 4
                )

[...]

PR #825, opened 2020-06-27
Resolves solariumphp/solarium#823 parse JsonRange facet into a JsonRange result

ee-dev:
Changes `Solarium\Component\ResponseParser\FacetSet::parseJsonFacetSet()`
The results of a `Solarium\Component\Facet\JsonRange` facet are parsed into a `Solarium\Component\Result\Facet\JsonRange` object. This is a subclass of `Solarium\Component\Facet\Buckets` which adds
```
getBefore()
getAfter()
getBetween()
```

Includes a new example 2.1.5.1.9-facet-json-range.php
Updated test Component/ResponseParser/FacetSetTest.php


src/Component/Result/Facet/JsonRange.php
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * @var int|null
+     */
+    protected $numBuckets;
+
+    /**
+     * Constructor.
+     *
+     * @param Bucket[] $buckets
+     * @param int|null $numBuckets
```
thomascorthals, line 55: You've probably overlooked this in a copy/paste. 😉

```suggestion
     * @param Bucket[] $buckets
     * @param int|null $before
     * @param int|null $after
     * @param int|null $between
```

ee-dev: Thanks, yep I used Buckets as a base and just realised I left in the other parent methods too! Will fix now…

examples/2.1.5.1.9-facet-json-range.php
```diff
@@ -0,0 +1,55 @@
...
+// get a select query instance
+$query = $client->createSelect();
+
+// get the facetset component
+$facetSet = $query->getFacetSet();
+
+// create a json range instance and set options
+// Set the 'other' parameter to retrieve before, after, between counts.
+// See https://lucene.apache.org/solr/guide/json-facet-api.html#range-facet-parameters
+$priceranges = new \Solarium\Component\Facet\JsonRange(['local_key' => 'priceranges', 'field' => 'price','start'=>1 ,'end'=>300,'gap'=>100,'other'=>'all']);
```
mkalkbrenner, line 18: In the example we should use the constant `JsonRange::OTHER_ALL` instead of 'all'.

tests/Component/ResponseParser/FacetSetTest.php
```diff
@@ -397,13 +397,49 @@ public function testParseJsonFacet()
...
+                        'count' => 2,
+                    ],
+                    'between' => [
+                        'count' => 14,
+                    ],
+                ],
             ],
         ];
 
+        $price_range = new \Solarium\Component\Facet\JsonRange(['local_key' => 'price_range', 'field' => 'price', 'start' => 1, 'end' => 300, 'gap' => 100, 'other' => 'all']);
```
mkalkbrenner, line 436: We should use the constant JsonRange::OTHER_ALL instead of 'all'.

examples/2.1.5.1.9-facet-json-range.php
ee-dev: Have switched to constants and listed possible values in the example

tests/Component/ResponseParser/FacetSetTest.php
ee-dev: changed to constant

ee-dev:
Really really sorry but I've just noticed I made an error in line 170 of Solarium\Component\ResponseParser\FacetSet
```
-                    if (isset($facets[$key]) && $facets[$key] instanceof JsonFacetInterface) {
+                    if (isset($facets[$key]) && $facets[$key] instanceof QueryFacetRange) {
```

As this PR is closed I think I need to create a new issue and PR?

PR #826, opened 2020-06-29
Update usage of implode

rodrigoaguilera:


thomascorthals:
But this was already fixed in the `master` branch for 5.1.5. We're on 6.0.0 now. If at all possible, I recommend you upgrade your Solarium version to get the most out of it.

mkalkbrenner:
Even if 3.x isn't supported anymore it doesn't hurt to commit this small fix.

rodrigoaguilera:
Thank you. This way we avoid warnings on php 7.4 :)

Issue #827, opened 2020-06-30
type checking error when parsing into a JsonRange


ee-dev:
All json facets returned as JsonRange instead of only json range facets.

PR #828, opened 2020-06-30
Fixes solariumphp/solarium#827

ee-dev:
Check facet type in the query to determine if a JsonRange result should be returned.


ee-dev:
The correct instanceof check is for 'QueryFacetJsonRange'

mkalkbrenner:
Create a new PR based on the current master.

PR #832, opened 2020-07-04
Facet term filtering options

### LINKED ISSUE #678: Support more domain options with JSON facet API
For instance, `excludeTags` seems unsupported:

https://lucene.apache.org/solr/guide/7_7/json-facet-api.html#filter-exclusions

ee-dev:
Added support for matches and excludeTerms on facet Field and FacetSet
```
Solarium\Component\FacetSet::setMatches()
Solarium\Component\FacetSet::setExcludeTerms()
Solarium\Component\Facet\Field::setMatches()
Solarium\Component\Facet\Field::setExcludeTerms()
```
Have included an example `2.1.5.1.1.1-facet-field-filters.php` which also shows the contains and prefix filters.
It might need a change to name/nesting if it doesn't fit in the hierarchy. I also removed the list of results from the example as it is just about facets but happy to put that back to match the other examples.

fixes #678


ee-dev:
StyleCI is complaining about a lack of period but it's a url. Is there a project specific way I should present the link to the SOLR issue that added the original feature?

wickedOne:
for links you can use the ``@see`` annotation as is done in the rest of the codebase.

out of curiousity: is there any specific reason you link to bug reports rather than documentation?

ee-dev:
> for links you can use the `@see` annotation as is done in the rest of the codebase.

Ah, I wasn't sure about using that as , as you note, it's a link to the issue patch not the standard documentation. 
> out of curiousity: is there any specific reason you link to bug reports rather than documentation?

There is an `@see` at the top of the class to the main documentation. but although the docs mention `excludeTerms` they don't explain it can be a list, whereas the implementation makes that clear. I added a similar link to `matches` so that the two were consistent. 
Happy to remove if you think they are superfluous, perhaps these simpler versions is all that is needed —
### For `setExcludeTerms()`
```* Exclude these terms, comma separated list. Use  \, for literal comma. Since Solr 6.5.```
### For `SetMatches()`
```* Limit facet terms to those matching this regular expression. Since Solr 7.2.```

wickedOne:
though i appreciate your consideration regarding consistency, i think adding the jira ticket link to ``setMatches`` might be somewhat confusing and the ``@see`` at the top of the class should be sufficient.

as for ``setExcludeTerms``: adding a comment on the method that it can be a list (as you did) is probably the best way to document it (not sure how many people go as far as reading the patches included in those solr's tickets).

so yes, please replace the links with your suggestions.
and be sure to add tests for the ``Solarium\Component\Facet\Field`` class as well

wickedOne:
fixes #678

ee-dev:
> fixes #678

Sorry, I don't think this is the same issue. This is for filtering 'SimpleFacets' not the newer Json API based facets.

ee-dev:
> though i appreciate your consideration regarding consistency, i think adding the jira ticket link to `setMatches` might be somewhat confusing and the `@see` at the top of the class should be sufficient.
> 
> as for `setExcludeTerms`: adding a comment on the method that it can be a list (as you did) is probably the best way to document it (not sure how many people go as far as reading the patches included in those solr's tickets).
> 
> so yes, please replace the links with your suggestions.
> and be sure to add tests for the `Solarium\Component\Facet\Field` class as well

Thanks. Have made the changes, seems clear now and you're right — no need to complicate things!

PR #835, opened 2020-07-07
use lucene-solr checkout

mkalkbrenner:


mkalkbrenner:
The reason for this PR is something different. In order to test more advanced stuff we need to include more java libs in solrconf.xml, for example the anlytics component.
If we keep our current approach, we need to maintain our own solrconf.xml for each Solr version. To avoid that I want to patch the included ones instead

mkalkbrenner:
Things are working now, except #864 

@wickedOne it should now be possible to add an integration test for the analytics component.
As #864 demonstrates, integretion tests are important ;-)

@thomascorthals So far we didn't document local testing well. If we merge this PR, we need to tell people to checkout lucene-solr. What do you think?

Issue #255, opened 2014-04-17
Implement schema API


basdenooijer:
Since 4.2 Solr has a schema API: https://wiki.apache.org/solr/SchemaRESTAPI

Solarium should support this.

bpolaszek:
Since Solr 5.0 the Schema API is more complete: 
https://cwiki.apache.org/confluence/display/solr/Schema+API

eostis:
Is there something new on the admin APIs of Solr ?

haydenyoung:
Is this still being considered for development? This would be a very useful feature especially when dealing with schemaless indexing.

mkalkbrenner:
We support the V1 CoreAdminAPI and the V1 CollectionsAPI partly. Other V1 APIs are not implemented.
But we added a generic implementation of the V2 API. Using it, you can do the V2 stuff mentioned in the docs:
https://lucene.apache.org/solr/guide/7_7/schema-api.html

See https://github.com/solariumphp/solarium/blob/bdb78b9260df9922161cfe8b4cf7df9243b50ccb/tests/Integration/AbstractTechproductsTest.php#L509 for an example

alexander-schranz:
Hello 👋 

I'm very new to Solr and I created a new collection. Via Solarium client, but how can I define the schema. In #839 it was mention there is already a way to define a schema via solarium, but I could not find anything in the docs? Maybe somebody can push me into the right direction :)

thomascorthals:
@alexander-schranz I'm going to answer in #1049 because I think this is not quite what you're looking for.

thomascorthals:
You can use a generic API query to update a schema through the [Schema API](https://solr.apache.org/guide/solr/latest/indexing-guide/schema-api.html).

https://github.com/solariumphp/solarium/blob/3e0eb01c8ef02554b5cc5755019704cce6b12d02/examples/execute_all.php#L101-L115

Issue #795, opened 2020-06-01
Use of 'final' and 'private' in Psr18Adapter


thomascorthals:
Is there any reason we can't use just `class` and `protected` for member variables and methods?

https://github.com/solariumphp/solarium/blob/559f8a8b4a6c7b764c404e544731b2226203c1c1/src/Core/Client/Adapter/Psr18Adapter.php#L16

The use of `final` and `private` contradicts our customisation philosophy.

https://github.com/solariumphp/solarium/blob/559f8a8b4a6c7b764c404e544731b2226203c1c1/docs/customizing-solarium.md#L8

https://github.com/solariumphp/solarium/blob/559f8a8b4a6c7b764c404e544731b2226203c1c1/docs/customizing-solarium.md#L299

mkalkbrenner:
See https://github.com/solariumphp/solarium/pull/754/files#r399572624


What is the reason for "final" (or the advantage)?

@dmaicher
I think we should not open this class for extension. Its does not make sense anyway as everything is private except public function execute(Request $request, Endpoint $endpoint): Response.

So we should not even encourage people to extend it. They should use composition if they need to customize stuff.

Or what do you think?

@mkalkbrenner
OK. Let's leave it like this and wait until someone complains ;-)

mkalkbrenner:
I accepted the proposal for a new implementation in 5.2.0 but I agree that we could be less restrictive for 6.0.0.
@thomascorthals @dmaicher please discuss your opionions :-)

thomascorthals:
I have missed that conversation about `final`. It happened to catch my eye because I was working on the docs. I have no idea how much sense it would make to extend it, it's more a matter of principle. If there's a good reason to make adapters `final`, we should state it in the docs. As far as unit and integration tests are concerned, it doesn't seem to matter.

A quick search of the code yields no other uses of `final` but quite a few `private`s. [CONTRIBUTING.md](https://github.com/solariumphp/solarium/blob/master/CONTRIBUTING.md) doesn't mention it specifically so it just might be that new contributors miss this entirely.

mkalkbrenner:
@wickedOne
@thomascorthals
@jsteggink
@thePanz
@timohund
@localheinz
@dmaicher

could we do a quick poll here?

localheinz:
@mkalkbrenner 

I am not a fan of inheritance and prefer composition instead, so for me, `final` classes (and consequently, `private` class members) work fine.

thomascorthals:
[...]

PR #839, opened 2020-07-10
solr v2 api manager

wickedOne:
@mkalkbrenner a quick draft just to see whether you like where this is going. 

keep in mind that this has been tested on solr non-cloud mode for the schema & config api only, so it's likely to need some additional work.

### a little background on some decisions
- the manager is final because of the poll @ https://github.com/solariumphp/solarium/issues/795
- the ``persist`` and ``flush`` is used because of familiarity. and while ``flush`` in reality is a ``reload`` i personally see no problem with that.
- i think solarium should only provide a ``ApiV2ResponseNormalizerInterface`` rather than actual normalizers (see below)
- the endpoint is optional in the constructor and a wither is provided to set it seperately as, for example, one might want to update multiple schema's in one iteration and re-instantiating a manager for that seems unnecessary.

### normalizers
writing your own normalizers should be easy and by not providing them out of the box, it keeps us from adding dependencies we don't really need. for instance a normalizer could look like:
```php
<?php

declare(strict_types=1);

namespace App\Manager;

use App\Model\MySchemaResponseModel;
use JMS\Serializer\SerializerInterface;
use Solarium\Core\Query\Result\ResultInterface;
use Solarium\Manager\Contract\ApiV2ResponseNormalizerInterface;

class SolrSchemaResponseNormalizer implements ApiV2ResponseNormalizerInterface
{
    /**
     * @var JMS\Serializer\SerializerInterface
     */
    private $serializer;

    /**
     * @param \JMS\Serializer\SerializerInterface $serializer
     */
    public function __construct(SerializerInterface $serializer)
    {
        $this->serializer = $serializer;
    }

    /**
     * {@inheritdoc}
     */
    public function normalize(ResultInterface $result)
    {
        return $this->serializer->deserialize($result->getBody(), MySchemaResponseModel::class, 'json');
    }
}
```

then creating the manager should be as simple as
```php
<?php

declare(strict_types=1);

namespace App;

use App\Manager\SolrSchemaResponseNormalizer;
use Solarium\Client;
use Solarium\Manager\Config\SchemaApiConfig;

class SolrSchemaService
{
    /**
     * @var \Solarium\Manager\ApiV2Manager
     */
    private $manager;

    /**
     * @param \Solarium\Client $client
     */
    public function __construct(Client $client)
    {
        $this->manager = new ApiV2Manager($client, new SchemaApiConfig(), new SolrSchemaResponseNormalizer());
    }
}
```


wickedOne:
perhaps we could provide a generic normalizer for those who don't care about the response?

wickedOne:
also, @dmaicher @thomascorthals @thePanz feel free to comment when you have any input on this..

thomascorthals:
The generic normalizer is a great idea.

If `flush` is actually a `reload`, maybe we should call it that for correctness's sake? We can still have `flush` as an alias for `reload`. Or perhaps the other way around if that would make more sense. As long as someone thinking "I need to reload" can do just that and have it work as expected.

wickedOne:
i initially chose ``flush`` because of familiarty for those using an orm like doctrine and because it's not clear to everyone that you need to reload the core in order to activate your changes.

call it ``reload`` and hava a ``flush`` method as alias indeed sounds like the best of both worlds.

thomascorthals:
This will also close #255.

PR #841, opened 2020-07-13
format date without modifying the value

### LINKED ISSUE #840: Helper::formatDate
https://github.com/solariumphp/solarium/blob/fe70c088f69c3b4fff35b8b5797e4be028af7044/src/Core/Query/Helper.php#L143

setting the timezone on an existing date time object, modifies the date time value when the original timezone differs:
https://3v4l.org/QPVOb

this will lead to unexpected results on each system not having UTC as the default timezone.
is this intended behaviour or a bug?

wouldn't it make more sense to simply use? 
```php
$dt->format('Y-m-d\TH:i:s\Z')
```

wickedOne:
fixes #840

### DIFF
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -115,14 +115,14 @@ class Helper
     {
         @trigger_error('The Helper::formatDate function is deprecated in Solarium 6.1 and will be removed in Solarium 6.2. Use Helper::convertAndFormatDate instead', E_USER_DEPRECATED);
 
-        return $this->convertAndFormatDate($input);
+        return $this->convertToUtcAndFormatDate($input);
     }
 
     /**
      * Convert and format a date to the expected formatting used in SOLR.
      *
      * Be aware this method adds a UTC TimeZone to the resulting date which may modify the date time value depending on
-     * the php configuration used running this code. Use Helper::dateToSolrUtcString to format your date without this
+     * the php configuration used running this code. Use Helper::formatDateAndPretendUtc to format your date without this
      * conversion.
      *
      * This format was derived to be standards compliant (ISO 8601)
@@ -135,7 +135,7 @@ class Helper
      *
      * @return string|bool false is returned in case of invalid input
      */
-    public function convertAndFormatDate($input)
+    public function convertToUtcAndFormatDate($input)
     {
         switch (true) {
             // input of datetime object
@@ -166,17 +166,21 @@ class Helper
         // when we get here the input is always a datetime object
         $input = $input->setTimezone(new \DateTimeZone('UTC'));
 
-        return $this->dateToSolrUtcString($input);
+        return $this->formatDateAndPretendUtc($input);
     }
 
     /**
      * Format date time interface to date string required by Solr.
      *
+     * Be aware this method just formats the date time as required by Solr without converting it into the UTC TimeZone.
+     * If your PHP application works with different time zones and doesn't convert to UTC before calling this function
+     * this might lead to unexpected errors when doing "date math". Consider Helper::convertToUtcAndFormatDate() instead.
+     *
      * @param \DateTimeInterface $dateTime
      *
      * @return string
      */
-    public function dateToSolrUtcString(\DateTimeInterface $dateTime): string
+    public function formatDateAndPretendUtc(\DateTimeInterface $dateTime): string
     {
         return $dateTime->format('Y-m-d\TH:i:s\Z');
     }
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -120,6 +120,7 @@ class Helper
             // input of datetime object
             case $input instanceof \DateTimeInterface:
                 $input = clone $input;
+
                 break;
             // input of timestamp or date/time string
             case \is_string($input):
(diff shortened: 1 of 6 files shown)


wickedOne:
i don't really get the coverage decrease of 0.01%...

src/Core/Query/Helper.php
```diff
@@ -101,7 +101,28 @@ public function escapePhrase(string $input): string
...
+    {
+        @trigger_error('The Helper::formatDate function is deprecated in Solarium 6.1 and will be removed in Solarium 6.2. Use Helper::convertAndFormatDate instead', E_USER_DEPRECATED);
+
+        return $this->convertAndFormatDate($input);
+    }
+
+    /**
+     * Convert and format a date to the expected formatting used in SOLR.
+     *
+     * Be aware this method adds a UTC TimeZone to the resulting date which may modify the date time value depending on
```
mkalkbrenner, line 123: Be aware that this method converts the input into UTC time zone as every dates have to be stored in UTC within Solr. So you might see that the date time "looks" different in the index and is "shifted" by some hours. But this is the correct way to handle different timezones in one index. But it is essential that you provide the correct TZ parameter when you do "date math" in a query. When you get a search result PHP could perfectly convert the date back into the time zone your application requires.
If you want to avoid that conversion you can use Helper::dateToSolrUtcString to format your date without it and just pretend that it is UTC.

src/Core/Query/Helper.php
```diff
@@ -113,52 +134,50 @@ public function escapePhrase(string $input): string
...
+
+        return $this->dateToSolrUtcString($input);
+    }
 
-        // unsupported input
-        return false;
+    /**
+     * Format date time interface to date string required by Solr.
+     *
+     * @param \DateTimeInterface $dateTime
```
mkalkbrenner, line 174: I would add some more explanations here:

Be aware this method just formats the date time as required by Solr without converting it into the UTC TimeZone. If your PHP application works with different time zones and doesn't convert to UTC before calling this function this might lead to unexpected errors when doing "date math". Consider Helper::convertToUtcAndFormatDate() instead.

src/Core/Query/Helper.php
```diff
@@ -113,52 +134,50 @@ public function escapePhrase(string $input): string
...
+     *
+     * @return string
+     */
+    public function dateToSolrUtcString(\DateTimeInterface $dateTime): string
```
mkalkbrenner, line 178: The function name sounds like a conversion to UTC. What about "formatDateAndPretendUtc".

src/Core/Query/Helper.php
```diff
@@ -113,52 +134,50 @@ public function escapePhrase(string $input): string
      *
      * @return string|bool false is returned in case of invalid input
      */
-    public function formatDate($input)
+    public function convertAndFormatDate($input)
```
mkalkbrenner, line 137: What about "convertToUtcAndFormatDate"?

src/Core/Query/Helper.php
wickedOne: yes, that's even more clear / better

src/Core/Query/Helper.php
wickedOne: yes, that's even more clear / better

src/Core/Query/Helper.php
wickedOne: yes, agreed. 

perhaps a little remark about the "out of sync" issue we've discussed before?
so basically one should use either method and not both depending on their workflow

wickedOne:
> We should a a dedicated page to the docs to explain the date time handling. I can do that when we agree on the function names.

yes, i'm fine with the proposed function names. 

a page dedicated to the fact solr only uses utc formatted date strings (apart from the documentation they provide themselves) and the different ways we provide of doing that might indeed be a good addition as i remember being puzzled by an unexpected search result using a date time filter in the past so hopefully will save some people from that headache

wickedOne:
i will not have time to work on this today though, might have tomorrow, else the earliest will be friday

thomascorthals:
> i don't really get the coverage decrease of 0.01%...

It's the -4 hits. There's also -4 lines that must have been hits before you removed them. Fewer hits in total for the codebase while your commit still has 100.0% coverage.

7388 / 8104 = 0.9116485686
7384 / 8100 = 0.9116049383

Rounding is mine. The difference between those is -0.00436303 with this rounding. I suspect they get rounded differently during the calculation. 0.91160 - 0.91165 is -0.005 which could then be rounded -0.01 for display. That would mean that the coverage for `master` and `#841` is either rounded from 0.911649 and 0.911605 or simply truncated.

At least, that's what I suspect is happening based on the numbers given and without looking in their source code.

> ```diff
> @@             Coverage Diff              @@
> ##             master     #841      +/-   ##
> ============================================
> - Coverage     91.16%   91.16%   -0.01%     
> - Complexity     3773     3774       +1     
> ============================================
>   Files           327      327              
>   Lines          8104     8100       -4     
> ============================================
> - Hits           7388     7384       -4     
>   Misses          716      716              
> ```

src/Core/Query/Helper.php
wickedOne: i'm not sure telling userland how to use dates in solr is helpful?

PR #842, opened 2020-07-14
Query builder

wickedOne:
helper class to help writing and maintaining (filter) queries; 
replaces https://github.com/solariumphp/solarium/pull/733

i had a look at the ``SearchApiSolrBackend::createFilterQuery`` method but concluded that the main difference between that class and this helper class is that the ``createFilterQuery`` method does a lot of "filling in the blanks" based on field types which is information not available in this class.

i did however add the ``Comparison::EMPTY`` constant and method to cover the edge case mentioned there twice for someone willing to query empty values. to keep things clean and simple, i've added a new comparison method rather than trying to incorporate it in others based on values passed.

also the addition two simple integrations in the query class ``Query::setQueryFromQueryBuilder`` and ``Query::addFilterQueriesFromQueryBuilder``.


docs/queries/query-helper/query-builder.md
```diff
@@ -0,0 +1,84 @@
...
+$value = 'foo:"bar"';
+$query->addFilterQuery(['key' => sha1($value), 'query' => $value]);
+$value = '-baz:"qux"';
+$query->addFilterQuery(['key' => sha1($value), 'query' => $value]);
+```
+
+Complex filter queries
+----------------------
+While the ``addFilterQueriesFromQueryBuilder`` method only provides in setting the facet query key and actual query, the ``QueryBuilder`` can be used in the construction of more complex facet queries.
+If one, for example, need to add a tag to the filter query the following method could be used.
```
thomascorthals, line 61:
```suggestion
If one, for example, needs to add a tag to the filter query the following method could be used.
```

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /**
+     * @param mixed  $value
+     * @param string $quote
+     * @param bool   $escape
+     *
+     * @return string
+     */
+    private function typedValueToString($value, string $quote = '', $escape = true): string
```
thomascorthals, line 179: One could theoretically wonder if `$escape` is nullable without a typehint.

```suggestion
    private function typedValueToString($value, string $quote = '', bool $escape = true): string
```

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,199 @@
...
+
+        foreach ($expr->getComparisons() as $child) {
+            $comparisons[] = $this->dispatch($child);
+        }
+
+        switch ($expr->getType()) {
+            case CompositeComparison::TYPE_AND:
+                return implode(' AND ', $comparisons);
+            case CompositeComparison::TYPE_OR:
+                return implode(' OR ', $comparisons);
```
thomascorthals, line 143: Should we use the constants `\Solarium\QueryType\Select\Query\Query::QUERY_OPERATOR_AND` and `QUERY_OPERATOR_OR` instead? It seems a bit overkillish and I don't think it has any benefits. Just wondering if you had considered it and what made you decide against it.

src/Builder/Select/QueryExpressionVisitor.php
```diff
@@ -0,0 +1,199 @@
...
+
+                return sprintf('%s:%s', $field, $this->valueToString($value, ',', '"'));
+            case Comparison::LIKE:
+            case Comparison::MATCH:
+                if (\is_array($value)) {
+                    return sprintf('%s:(%s)', $field, $this->valueToString($value, ' OR ', '', false));
+                }
+
+                return sprintf('%s:%s', $field, $this->valueToString($value, ',', '', false));
+            case Comparison::NIN:
```
thomascorthals, line 99: I do have a strong opinion about the order of the cases. I think `NIN` should come right after `IN`.

Personally, I'd stick to the same order for all the cases as the `const`s in `Comparison` are defined in.

src/Builder/Select/ExpressionBuilder.php
```diff
@@ -0,0 +1,189 @@
...
+     * @param null $x
+     *
+     * @throws \Solarium\Exception\RuntimeException
+     *
+     * @return \Solarium\Builder\CompositeComparison
+     */
+    public function andX($x = null): CompositeComparison
+    {
+        return new CompositeComparison(CompositeComparison::TYPE_AND, \func_get_args());
+    }
```
thomascorthals, line 34: The [`...` operator](https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list) makes this typehintable and thus more readable.

```suggestion
    /**
     * @param Comparison ...$comparison
     *
     * @throws \Solarium\Exception\RuntimeException
     *
     * @return \Solarium\Builder\CompositeComparison
     */
    public function andX(Comparison ...$comparison): CompositeComparison
    {
        return new CompositeComparison(CompositeComparison::TYPE_AND, $comparison);
    }
```

src/Builder/Select/ExpressionBuilder.php
```diff
@@ -0,0 +1,189 @@
...
+     * @param null $x
+     *
+     * @throws \Solarium\Exception\RuntimeException
+     *
+     * @return \Solarium\Builder\CompositeComparison
+     */
+    public function orX($x = null): CompositeComparison
+    {
+        return new CompositeComparison(CompositeComparison::TYPE_OR, \func_get_args());
+    }
```
thomascorthals, line 46:
```suggestion
    /**
     * @param Comparison ...$comparison
     *
     * @throws \Solarium\Exception\RuntimeException
     *
     * @return \Solarium\Builder\CompositeComparison
     */
    public function orX(Comparison ...$comparison): CompositeComparison
    {
        return new CompositeComparison(CompositeComparison::TYPE_OR, $comparison);
    }
```

thomascorthals (changes requested):
Looks good as a whole. My remarks are mostly about readability rather than functionality.

thomascorthals:
Sorry for messing that up as multiple comments instead of a single review!

wickedOne:
@thomascorthals thanx for your review. i do agree with all of your suggestions.

as for the usage of constants: initially this was written to be a completely seperate component outside of the ``Solarium`` namespace thus no dependecies and it can be considered a leftover. 

nevertheless i do prefer to have my constants defined in the namespace handling them for clarity. 
but i'm fine either way...

thomascorthals:
@wickedOne On the other hand, I don't think a hardcoded `AND` and `OR` will get us into trouble with Solr. If they change _that_ syntax, we're probably in for one hell of a ride.

tests/Builder/Select/QueryBuilderTest.php
```diff
@@ -0,0 +1,369 @@
...
+    {
+        $filter = QueryBuilder::create()
+            ->where(QueryBuilder::expr()->notIn('foo', [2, 5]));
+
+        $this->assertSame('-foo:(2 OR 5)', $this->visitor->dispatch($filter->getExpressions()[0]));
+
+        $filter = QueryBuilder::create()
+            ->where(QueryBuilder::expr()->notIn('foo', 'bar'));
+
+        $this->assertSame('-foo:"bar"', $this->visitor->dispatch($filter->getExpressions()[0]));
```
mkalkbrenner, line 192: In most cases you want `*:* -foo:"bar"`here.

src/Builder/Select/ExpressionBuilder.php
wickedOne: i've been working quite a bit with the [FunctionBuilder](https://github.com/solariumphp/solarium/blob/master/src/Builder/Analytics/FunctionBuilder.php) recently and one of the benefits of not typehinting the arguments is that you can pass a string as argument for the edge cases you run into.

for instance creating this function [``filter(mult(price,quantity),or(cats:equal(category,_)))``](https://lucene.apache.org/solr/guide/analytics.html#for-each-functions) would translate into
```php
$expression = FunctionBuilder::create()
            ->where($expr->filter($expr->mult('price', 'quantity'), $expr->or('cats:equal(category,_)')))
        ;
```
as ``$expr->equals`` would render a different expression, it cannot be used to generate the desired result.

as this is a helper class, i'd personally prefer to make it not too strict; what do you think?

wickedOne:
> We need to mention that this query builder must only be used in combination with the standard query parser (lucene) or edismax.

i think there's quite a few query parsers which also work with the standard query language?

> The default operator is "OR". The resulting string will be shorter if we use a single space as glue instead of " OR ".
And I prefer to do +a:1 +b:2instead of a:1 AND b:1.
This way you can reduce the length of the query string. This way you could avoid switching to POST requests more often and keep GET. For POST requests some of Solr's caches are disabled.

fine with me, but are you sure that also works with the Lucene, Payload Score or Surround Query Parser to name a few?
perhaps better to instantiate the builder with the operators you want?

> In general I consider the query builder to be useful. But we need to add documentation. Having a query with much boolean operators is slow. In most cases it makes a lot of sense to split the query in multiple filter queries which could be cached individually. So the examples should consist of code where we use this query builder to create a query in combination with multiple filter queries.

you think this is the responsibility of the solarium library? if so, we might want to consider a "best practices" section in the documentation (which i think would be very helpful) as these considerations are not limited to using a query builder

mkalkbrenner:
> > We need to mention that this query builder must only be used in combination with the standard query parser (lucene) or edismax.
> 
> i think there's quite a few query parsers which also work with the standard query language?

Yes, but not all.

> > The default operator is "OR". The resulting string will be shorter if we use a single space as glue instead of " OR ". And I prefer to do +a:1 +b:2instead of a:1 AND b:1. This way you can reduce the length of the query string. This way you could avoid switching to POST requests more often and keep GET. For POST requests some of Solr's caches are disabled.
> 
> fine with me, but are you sure that also works with the Lucene, Payload Score or Surround Query Parser to name a few?
> perhaps better to instantiate the builder with the operators you want?

It is the opposite. AND and OR are not accepted by all parsers. See https://lucene.apache.org/solr/guide/7_3/the-standard-query-parser.html#the-standard-query-parser

**The standard query parser supports all the Boolean operators listed in the table above. The DisMax query parser supports only + and -.** 

> > In general I consider the query builder to be useful. But we need to add documentation. Having a query with much boolean operators is slow. In most cases it makes a lot of sense to split the query in multiple filter queries which could be cached individually. So the examples should consist of code where we use this query builder to create a query in combination with multiple filter queries.
> 
> you think this is the responsibility of the solarium library? if so, we might want to consider a "best practices" section in the documentation (which i think would be very helpful) as these considerations are not limited to using a query builder

You're right.

wickedOne:
would you agree to a solution which by default uses + & - as operators which can be turned into AND and OR by using a flag on instantiation?

mkalkbrenner:
Sure!

PR #844, opened 2020-07-14
bug report suggestion

wickedOne:
bug report template suggestion to get all the info to quickly reproduce. 

not sure whether we initially need more info like platform running solr, active components, solr configs, etc.


wickedOne:
ugh, apparently i forgot to click save somewhere :|

PR #846, opened 2020-07-15
symfony coding standard

### LINKED ISSUE #747: Coding style adjustments?

wickedOne:
fixes #747 

```
vendor/bin/phpcs
vendor/bin/php-cs-fixer fix
```
can now be run without resulting in warnings / errors / modifications

### remarks:
- i've changed all events interfaces to not instantiable classes so there's no need for users of those to do any refactoring. nevertheless, do you want me to make an entry in the changelog for that?
- i'm a bit puzzled by style ci; though they offer a symfony preset, it clearly does not apply all rules described at [symfony's coding standard page](https://symfony.com/doc/current/contributing/code/standards.html) so how to go forward from here: do we move away from ctyle ci or do we add an additional style check step; any suggestions?

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -1,5 +1,12 @@
 <?php
 
+/*
+ * This file is part of the Solarium package.
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
+ */
+
 namespace Solarium;
 
 use Solarium\Core\Client\Client as CoreClient;
@@ -66,7 +73,7 @@ class Client extends CoreClient
      */
     public static function checkExact(string $version): bool
     {
-        return substr(self::VERSION, 0, strlen($version)) == $version;
+        return 0 === strpos(self::VERSION, $version);
     }
 
     /**
--- a/src/Component/Facet/JsonFacetTrait.php
+++ b/src/Component/Facet/JsonFacetTrait.php
@@ -1,9 +1,17 @@
 <?php
 
+/*
+ * This file is part of the Solarium package.
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
+ */
+
 namespace Solarium\Component\Facet;
 
 use Solarium\Component\FacetSetInterface;
 use Solarium\Component\FacetSetTrait;
+use Solarium\Core\Configurable;
 use Solarium\Core\Query\Helper;
 use Solarium\Exception\InvalidArgumentException;
 
@@ -40,6 +48,7 @@ trait JsonFacetTrait
     public function getDomainFilter()
     {
         $domain = $this->getOption('domain');
+
         return $domain['filter'] ?? null;
     }
 
@@ -51,9 +60,9 @@ trait JsonFacetTrait
      * @param string $query
      * @param array  $bind  Bind values for placeholders in the query string
      *
-     * @return self Provides fluent interface
+     * @return \Solarium\Core\Configurable
      */
-    public function setDomainFilterQuery(string $query, array $bind = null): FacetSetInterface
+    public function setDomainFilterQuery(string $query, array $bind = null): Configurable
     {
         if (null !== $bind) {
             $helper = new Helper();
@@ -61,20 +70,22 @@ trait JsonFacetTrait
         }
 
         $filter = $this->getDomainFilter();
-        if (!$filter || is_string($filter)) {
+        if (!$filter || \is_string($filter)) {
             return $this->setOption('domain', ['filter' => $query]);
         }
 
-        foreach ($filter as &$param_or_query) {
-            if (is_string($param_or_query)) {
-                $param_or_query = $query;
+        foreach ($filter as &$paramOrQuery) {
+            if (\is_string($paramOrQuery)) {
+                $paramOrQuery = $query;
+
                 return $this->setOption('domain', ['filter' => $filter]);
             }
         }
-        unset($param_or_query);
+        unset($paramOrQuery);
 
         /* @noinspection UnsupportedStringOffsetOperationsInspection */
         $filter[] = $query;
+
         return $this->setOption('domain', ['filter' => $filter]);
     }
 
@@ -92,19 +103,20 @@ trait JsonFacetTrait
             return $this->setOption('domain', ['filter' => ['param' => $param]]);
         }
 
-        if (is_string($filter) || 1 == count($filter)) {
+        if (\is_string($filter) || 1 === \count($filter)) {
             return $this->setOption('domain', ['filter' => [$filter, ['param' => $param]]]);
         }
 
-        foreach ($filter as &$param_or_query) {
-            if (is_array($param_or_query) && $param_or_query['param'] == $param) {
+        foreach ($filter as &$paramOrQuery) {
+            if (\is_array($paramOrQuery) && $paramOrQuery['param'] === $param) {
                 return $this;
             }
         }
-        unset($param_or_query);
+        unset($paramOrQuery);
 
         /* @noinspection UnsupportedStringOffsetOperationsInspection */
         $filter[] = ['param' => $param];
+
         return $this->setOption('domain', ['filter' => $filter]);
     }
 
@@ -137,7 +149,6 @@ trait JsonFacetTrait
 
     /**
      * Add a facet.
-     *
      *
      * @param FacetInterface|array $facet
      *
--- a/src/Component/Facet/JsonFacetTrait.php
+++ b/src/Component/Facet/JsonFacetTrait.php
@@ -11,7 +11,6 @@ namespace Solarium\Component\Facet;
 
 use Solarium\Component\FacetSetInterface;
 use Solarium\Component\FacetSetTrait;
-use Solarium\Core\Configurable;
 use Solarium\Core\Query\Helper;
 use Solarium\Exception\InvalidArgumentException;
 
@@ -60,9 +59,9 @@ trait JsonFacetTrait
      * @param string $query
      * @param array  $bind  Bind values for placeholders in the query string
      *
-     * @return \Solarium\Core\Configurable
+     * @return \Solarium\Component\FacetSetInterface
      */
-    public function setDomainFilterQuery(string $query, array $bind = null): Configurable
+    public function setDomainFilterQuery(string $query, array $bind = null): FacetSetInterface
     {
         if (null !== $bind) {
             $helper = new Helper();
--- a/src/Core/Client/Response.php
+++ b/src/Core/Client/Response.php
@@ -1,5 +1,12 @@
 <?php
 
+/*
+ * This file is part of the Solarium package.
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
+ */
+
 namespace Solarium\Core\Client;
 
 use Solarium\Exception\HttpException;
@@ -94,12 +101,11 @@ class Response
     /**
      * Set headers.
      *
-     *
      * @param array $headers
      *
-     * @return self Provides fluent interface
-     *
      * @throws HttpException
+     *
+     * @return self Provides fluent interface
      */
     public function setHeaders(array $headers): self
     {
@@ -108,7 +114,7 @@ class Response
         // get the status header
         $statusHeader = null;
         foreach ($headers as $header) {
-            if ('HTTP' == substr($header, 0, 4)) {
+            if (0 === strpos($header, 'HTTP')) {
                 $statusHeader = $header;
                 break;
             }
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -1,5 +1,12 @@
 <?php
 
+/*
+ * This file is part of the Solarium package.
+ *
+ * For the full copyright and license information, please view the COPYING
+ * file that was distributed with this source code.
+ */
+
 namespace Solarium\Core\Query;
 
 use Solarium\Exception\InvalidArgumentException;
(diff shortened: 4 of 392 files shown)


dmaicher:
I would also vote to add `'no_superfluous_phpdoc_tags' => true,` to the php-cs-fixer config. 

This would remove redundant/superfluous `@param` phpdoc tags.

See https://symfony.com/doc/current/contributing/code/standards.html#symfony-coding-standards-in-detail

wickedOne:
regarding the superflous phpdoc tags:
```
find "$(pwd)" -name Query.php
solarium/src/Component/Facet/Query.php
solarium/src/Component/Result/Facet/Query.php
solarium/src/Plugin/MinimumScoreFilter/Query.php
solarium/src/QueryType/Graph/Query.php
solarium/src/QueryType/Terms/Query.php
solarium/src/QueryType/Spellcheck/Query.php
solarium/src/QueryType/Select/Query/Query.php
solarium/src/QueryType/Ping/Query.php
solarium/src/QueryType/Suggester/Query.php
solarium/src/QueryType/Server/Collections/Query/Query.php
solarium/src/QueryType/Server/CoreAdmin/Query/Query.php
solarium/src/QueryType/Server/Api/Query.php
solarium/src/QueryType/MoreLikeThis/Query.php
solarium/src/QueryType/RealtimeGet/Query.php
solarium/src/QueryType/Stream/Query.php
solarium/src/QueryType/Extract/Query.php
solarium/src/QueryType/Update/Query/Query.php
```

so what is the benefit of having
```php
public function myFunction(Query $query): void
```
instead of
```php
/**
 * @param Solarium\QueryType\Stream\Query $query
 */
public function myFunction(Query $query): void
```
?

for me this is essential when doing code reviews. apart from that, having one function with, and the other without a phpdoc block is super inconsistent and uneasy to "scan" when doing code reviews so to me personally that is the added value. 

> Add PHPDoc blocks for all classes, methods, and functions (though you may be asked to remove PHPDoc that do not add value);

https://symfony.com/doc/current/contributing/code/standards.html#documentation

thomascorthals:
Personally, I also find it very confusing that a method with two parameters only has one `@param`. It somehow looks like the developer didn't bother to make the PHPdoc match the code.

```php
    /**
     * @param string $dummy Some argument description
     */
    public function __construct($dummy, Qux $qux)
    {
        $this->fooBar = $this->transformText($dummy);
        $this->qux = $qux;
    }
```

dmaicher:
Maybe its just me then who dislikes redundant phpdoc :blush: 

- more lines of code to read and review
- more overhead when writing code: need to make phpdoc "match the code"
- more error prone: phpdoc might not reflect the actual code

Anyway this PR is a great step towards more consistency :+1: 

But if we keep `no_superfluous_phpdoc_tags = false` then we should enforce phpdoc tags for all arguments and return types? Currently its not consistent.

thomascorthals:
I totally agree that having documentation (PHPdoc and user manual) that tells something different than what's actually happening is worse than having no documentation at all.

In the case of redundant PHPdoc, it's objectively not wrong to omit it but I find it subjectively "less right". It's not really a problem when reviewing. If you add a parameter without PHPdoc and I'm reviewing, I'll see what it is doing in the code and grasp what it's supposed to be doing from your commit messages and PR description. But if I'm coming across existing code where the last parameter is omitted from the PHPdoc, I'll get suspicious and wonder if whoever added that parameter might have forgotten to update the PHPdoc.

An additional boolean, for instance, might mean some new behaviour is now possible for the function upon request, _or_ the default behaviour for the function has changed and the old behaviour is still available upon request. I'd be more confident in the remainder of the PHPdoc if that boolean is explicitly listed because then I know it wasn't overlooked when updating the code. Ideally, the function description should clarify this further of course and leave nothing to be guessed.

We should consider what's the least effort for our users before we consider what's the least effort for us.

- You write the code once.
- It is reviewed by maybe 2 or 3 people who read your PR and can easily reach you for clarifications.
- It will be read and perhaps updated by dozens of coming and going developers who have to figure it out when you might no longer be actively involved.
- For "user facing functions", it will be used by hundreds or thousands of developers of varying backgrounds and experience. You can't assume they read `CONTRIBUTING` or are familiar with the Symfony CS. They might not know omitted PHPdoc is intentional and carries a meaning in and of itself. And their IDE might only do typehints from PHPdoc and not yet from the fairly new PHP syntax.

In my opinion, the whole point of a library is that we put in the effort so our users don't have to. And those users are a homogeneous group. The question I think we should be asking is: "Is it better to mildly annoy users who find it redundant, or to confuse users who are left guessing?"

Keeping the users in mind, we should probably enforce PHPdoc for all arguments and return types of `public` functions.

It looks like my 2 cents turned out te be 2 euros. 😉

mkalkbrenner:
We had the same discussion some time ago ;-)

I agree that
"we should ~~probably~~ enforce PHPdoc for all arguments and return types of public functions."

mkalkbrenner:
@wickedOne Could we quickly solve the merge conflicts?
I'll block all other PRs until this one gets committed to avoid further conflicts.
The doc blocks could optionally be completed in a second PR.

src/Component/Facet/JsonFacetTrait.php
```diff
@@ -51,30 +60,32 @@ public function getDomainFilter()
      * @param string $query
      * @param array  $bind  Bind values for placeholders in the query string
      *
-     * @return self Provides fluent interface
+     * @return \Solarium\Core\Configurable
      */
-    public function setDomainFilterQuery(string $query, array $bind = null): FacetSetInterface
+    public function setDomainFilterQuery(string $query, array $bind = null): Configurable
```
mkalkbrenner, line 65: I know that our interfaces aren't perfect. But returning FacetSetInterface might be better here.

wickedOne:
@mkalkbrenner sorry for the late reply, had some private stuff which needed to be dealt with first...

conflicts resolved, return type fixed

thomascorthals:
Looks good at first sight.

I had those `sprintf`s for exceptions lined up too. Was just holding back on the PR until other stuff got merged. 😅

src/Client.php
```diff
@@ -66,7 +73,7 @@ class Client extends CoreClient
      */
     public static function checkExact(string $version): bool
     {
-        return substr(self::VERSION, 0, strlen($version)) == $version;
+        return 0 === strpos(self::VERSION, $version);
```
thomascorthals, line 76:
```suggestion
        return 0 === \strpos(self::VERSION, $version);
```

src/Core/Client/Response.php
```diff
@@ -108,7 +114,7 @@ public function setHeaders(array $headers): self
         // get the status header
         $statusHeader = null;
         foreach ($headers as $header) {
-            if ('HTTP' == substr($header, 0, 4)) {
+            if (0 === strpos($header, 'HTTP')) {
```
thomascorthals, line 117:
```suggestion
            if (0 === \strpos($header, 'HTTP')) {
```

src/Core/Query/Helper.php
```diff
@@ -114,10 +121,10 @@ public function formatDate($input)
                 $input = clone $input;
                 break;
             // input of timestamp or date/time string
-            case is_string($input) || is_numeric($input):
-
+            case \is_string($input):
+            case is_numeric($input):
```
thomascorthals, line 125:
```suggestion
            case \is_numeric($input):
```

src/QueryType/ManagedResources/Result/Resources/Resource.php
```diff
@@ -105,9 +118,9 @@ public function setClass(string $class): self
      */
     public function getType(): string
     {
-        if (0 === strncmp($this->resourceId, '/schema/analysis/stopwords', strlen('/schema/analysis/stopwords'))) {
+        if (0 === strncmp($this->resourceId, '/schema/analysis/stopwords', \strlen('/schema/analysis/stopwords'))) {
```
thomascorthals, line 121:
```suggestion
        if (0 === \strncmp($this->resourceId, '/schema/analysis/stopwords', \strlen('/schema/analysis/stopwords'))) {
```

src/QueryType/ManagedResources/Result/Resources/Resource.php
```diff
@@ -105,9 +118,9 @@ public function setClass(string $class): self
...
             return self::TYPE_STOPWORDS;
-        } elseif (0 === strncmp($this->resourceId, '/schema/analysis/synonyms', strlen('/schema/analysis/synonyms'))) {
+        } elseif (0 === strncmp($this->resourceId, '/schema/analysis/synonyms', \strlen('/schema/analysis/synonyms'))) {
```
thomascorthals, line 123:
```suggestion
        } elseif (0 === \strncmp($this->resourceId, '/schema/analysis/synonyms', \strlen('/schema/analysis/synonyms'))) {
```

src/QueryType/Stream/ExpressionBuilder.php
```diff
@@ -244,17 +251,17 @@ class ExpressionBuilder
...
-     *
      * @throws InvalidArgumentException
+     *
+     * @return string
      */
     public function __call(string $name, array $arguments): string
     {
         return $name.'('.implode(', ', array_filter($arguments, function ($value) {
-            if (is_array($value) || (is_object($value) && !method_exists($value, '__toString'))) {
+            if (\is_array($value) || (\is_object($value) && !method_exists($value, '__toString'))) {
```
thomascorthals, line 261:
```suggestion
            if (\is_array($value) || (\is_object($value) && !\method_exists($value, '__toString'))) {
```

src/QueryType/Stream/ExpressionBuilder.php
```diff
@@ -244,17 +251,17 @@ class ExpressionBuilder
...
                 throw new InvalidArgumentException('An expression argument must be a scalar value or an object that provides a __toString() method.');
             }
-            if (is_string($value)) {
+            if (\is_string($value)) {
                 $value = trim($value);
```
thomascorthals, line 265:
```suggestion
                $value = \trim($value);
```

src/QueryType/Stream/ExpressionBuilder.php
```diff
@@ -271,29 +278,30 @@ public function __call(string $name, array $arguments): string
...
+        $indentedExpression = '';
+        $len = \strlen($expression);
         for ($c = 0; $c < $len; ++$c) {
             if ('(' === $expression[$c]) {
-                $indented_expression .= $expression[$c].PHP_EOL;
-                $current_indentation += $indentation_step;
-                $indented_expression .= str_pad('', $current_indentation);
+                $indentedExpression .= $expression[$c].PHP_EOL;
+                $currentIndentation += $indentationStep;
+                $indentedExpression .= str_pad('', $currentIndentation);
```
thomascorthals, line 289:
```suggestion
                $indentedExpression .= \str_pad('', $currentIndentation);
```

src/QueryType/Stream/ExpressionBuilder.php
```diff
@@ -271,29 +278,30 @@ public function __call(string $name, array $arguments): string
...
             } elseif (')' === $expression[$c]) {
-                $current_indentation -= $indentation_step;
-                $indented_expression .= PHP_EOL;
-                $indented_expression .= str_pad('', $current_indentation).$expression[$c];
+                $currentIndentation -= $indentationStep;
+                $indentedExpression .= PHP_EOL;
+                $indentedExpression .= str_pad('', $currentIndentation).$expression[$c];
```
thomascorthals, line 293:
```suggestion
                $indentedExpression .= \str_pad('', $currentIndentation).$expression[$c];
```

src/QueryType/Stream/ExpressionBuilder.php
```diff
@@ -271,29 +278,30 @@ public function __call(string $name, array $arguments): string
...
             } elseif (',' === $expression[$c]) {
-                $indented_expression .= $expression[$c].PHP_EOL.str_pad('', $current_indentation);
+                $indentedExpression .= $expression[$c].PHP_EOL.str_pad('', $currentIndentation);
```
thomascorthals, line 295:
```suggestion
                $indentedExpression .= $expression[$c].PHP_EOL.\str_pad('', $currentIndentation);
```

src/Support/DataFixtures/FixtureLoader.php
```diff
@@ -42,9 +49,9 @@ public function __construct(Loader $loader, Purger $purger, Executor $executor)
      * @param string $dir
      * @param bool   $append
      *
-     * @return self
-     *
      * @throws ReflectionException
+     *
+     * @return self
      */
     public function loadFixturesFromDir(string $dir, bool $append = true)
```
thomascorthals, line 56:
```suggestion
    public function loadFixturesFromDir(string $dir, bool $append = true): self
```

wickedOne:
as far as i know prefixing with a ``\`` is done to improve opcode caching for certain functions.

those prefixed are listed here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3e5dd53a0ab1fbe87057b06a027d8d072457ca3e/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L355-L396

prefixing sprintf and other would result in php-cs-fixer changing them back to non prefixed

thomascorthals:
My bad. I failed to see a connection between them, other than language constructs were obviously excluded. You learn something every day. 🤓

wickedOne:
yes, and i'm glad we do 😄 

fixed the return type and added several others to the mix (although it's not part of the coding standard, it's a good thing to have)

PR #847, opened 2020-07-15
Refactor duplicate code in Managed Resources

thomascorthals:
I've tried to get rid of duplicate code in Managed Resources by putting shared functionality in abstract classes. Most of the `Command`s are even identical for stopwords and synonyms.

My intention was to keep full backward compatibility. I ended up comprimising a little bit. It's currently no longer possible to instantiate most commands like `new \Solarium\QueryType\ManagedResources\Query\Stopwords\Command\Remove()` because they are no longer in the separate namespaces. Only `Add` and `Create` are still available like that. But `$query->createCommand(Stopwords::REMOVE)` keeps working as expected without code changes for the user. The former behaviour was used in the integration tests, but it wasn't 'officially' documented anywhere. Neither is the latter at the moment, but it's in the same spirit as `$client->createSelect()` et al.

I see three possibilities regarding BC, and will be happy to update this PR if 2 or 3 is preferred to 1.

1. Assume `createCommand()` is The Right Way to do it because it hides the implementation details from the library user. Add a note in the changelog to this effect and make it clear in the documentation when we get around to writing it.
2. Keep the constructors in the separate namespaces working for now by extending each command in both namespaces. The extending classes would be empty and marked `@deprecated`.
3. Keep them available in the separate namespaces in perpetuity. In that case, I would make all commands under `Query\Command` abstract. The extending classes would still be empty. This would be a clear sign the distinction was made intetionally.

For good measure, I threw in `createInitArgs()` to hide those implementation details as well.


thomascorthals:
I pushed a commit to the branch on my repo to fix a style issue and it isn't showing up in this PR. Did I break GitHub?

wickedOne:
i agree with the ``createCommand`` and ``createInitArgs`` approach (1) as it's consistent with the majority of the codebase while it provides the ones working on the codebase with the most flexibility

thomascorthals:
If we do stick with (1), which makes more sense?

(a) `Query\Stopwords\Command\Add.php`

—or—

(b) `Query\Command\Stopwords\Add.php`

I left it at (a) because that's how it was. But maybe it's better to have everything under `Query\Command` for us internally? Doesn't matter for the user who goes through `createCommand`.

wickedOne:
i'd opt for the ``Query\Command`` namespace indeed

thomascorthals:
You can pass your init args directly to `createInitArgs` now.

```php
$query = $client->createManagedStopwords();
$initArgs = $query->createInitArgs(['ignoreCase' => true]);
```

mkalkbrenner:
> 1. Assume createCommand() is The Right Way to do it because it hides the implementation details from the library user. Add a note in the changelog to this effect and make it clear in the documentation when we get around to writing it.

I agree with option 1, too. So the remaining tasks are to complete the documentation and to update the CHANGELOG before we can merge this PR.

thomascorthals:
@mkalkbrenner 'Completing' is a bit of a euphemism for documentation that's historically at 0%. 😉

I don't mind writing the documentation some day, but is it really necessary to include it in this PR?

mkalkbrenner:
@thomascorthals no ;-)
But the required minimum is the adjusted CHANGELOG and to open an issue for the documentation task.

thomascorthals:
I've opened #848 for the documentation task.

PR #850, opened 2020-07-29
Add `getWasSuccessful` to Managed Resources command results

### LINKED ISSUE #830: Getting a single managed stopword/synonym
We're currently doing this
```php
$query = $client->createManagedSynonyms();
$query->setName('english');
$result = $client->execute($query);
```
to get the result for
```bash
curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/english"
```

I'm looking to implement support for getting a single synonym mapping (or stopword) as in
```bash
curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/english/happy"
```

There are two ways I can go about this.

1. Set a term on the query.
```php
$query = $client->createManagedSynonyms();
$query->setName('english');
$query->setTerm('happy');
```

Pro: this is the simplest way of setting a term from a user's point of view.
Con: the result would be of a different type with or without the term set on the query.

2. Add a `Get` action.
```php
$query = $client->createManagedSynonyms();
$query->setName('english');
$get = new Get();
$get->setTerm('happy');
$query->setCommand($get);
```

Pro: one type of result for one type of command.
Con: depending on what you want to `GET` (as an HTTP verb), you do or don't need to set a command.

I think a separate `Get` command is the best choice. Unless there's a good reason I'm missing to go with the first option?

I'm also wondering what to do when the term doesn't exist as a synonym/stopword. Solr will respond with a 404. Throwing a `RuntimeException` isn't really wrong but might be a bit harsh.

There's also the `Exists` command. You have to check the status code against 200 or 404 to know if the term exists. The implementation is based on this info from the [ref guide](https://lucene.apache.org/solr/guide/8_5/managed-resources.html):
> You can test to see if a specific word exists by sending a GET request for that word as a child resource of the set, such as:
> ```bash
> curl "http://localhost:8983/solr/techproducts/schema/analysis/stopwords/english/foo"
> ```
> This request will return a status code of 200 if the child resource (foo) exists or 404 if it does not exist the managed list. 

It happens to also return the data I want from `Get`. But you can't tell an end user to use `Exists` to "get" something and parse it yourself. In fact, as a library user I would expect to get a boolean reply to the question if something exists.

Any thoughts and suggestions are appreciated. I want to implement this first and then use it in the integration test for #742.

thomascorthals:
Instead of having to check if the status code was 200, you can now call `getWasSuccessful()` after executing a command on a managed resource. I had to introduce a separate response parser and result class as the response doesn't look anything like what you get for requesting resources.

This is only half of the work to make the `Exists` command actually useful. I still have to deal with the scenario where the command isn't successful. It currently throws an exception when a boolean `false` is what a user would reasonably expect.

Since this doesn't break anything, I prefer to have it reviewed early on to be sure I'm on the right track.

EDIT: With the latest commits, this now closes #830.


thomascorthals:
The Style CI failure doesn't seem to be caused by this PR. I tried fixing it by removing `blank_line_before_break` from the `disabled` section in `.styleci.yml` but then  it would complain about the next line and if I remove that it would complain about the next next one …

So I reverted those to keep my PR about what I actually added.

wickedOne:
i haven't tested this pr yet, but is it about parsing a response similar to this?

Screenshot 2020-03-10 at 14 08 16

which gives "weird" responses because of
Screenshot 2020-03-10 at 14 28 13

if so: i'd opt for a more generic handling of solr errors perhaps in the ``AbstractResponseParser``. for example having an error object with correct metadata, msg & code available when solr responds with anything other than a 200 response

if not: i'll have a closer look at this over the weekend and will open a new ticket for better solr error handling

in general: you might want to consider to rename ``getWasSuccessful`` to ``isSuccessful`` as issers appears to be the new standard when returning booleans.

thomascorthals:
@wickedOne Mostly, but there's also a use case I'm facing where I want to find out if a stopword/synonym list exists as a whole instead of finding out if a term exists in it.

```bash
$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/english/nonexistant"
{
  "responseHeader":{
    "status":404,
    "QTime":0},
  "error":{
    "metadata":[
    "msg":"nonexistant not found in /schema/analysis/synonyms/english",
    "code":404}}

$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/engrish"
<html>
<head>
   <title>Status page</title>
</head>
<body style="font-family: sans-serif;">
<p style="font-size: 1.2em;font-weight: bold;margin: 1em 0px;">Not Found</p>
<p>No REST managed resource registered for path /schema/analysis/synonyms/engrish</p>
<p>You can get technical details <a href="http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.5">here</a>.<br>
Please continue your visit at our <a href="/">home page</a>.
</p>
</body>
</html>
```

Both have a `HTTP/1.1 404 Not Found` response header. I want to be able to handle the second case as gracefully as the first one from a user viewpoint. We're here to do the heavy lifting from a library viewpoint.

Anyway, feedback like yours is the reason I opened this PR before I went any further with it!

I went with `getWasSuccessful` because it was used in a handful of other places in Solarium. I agree in principle with `isSuccessful`. It would be inconsistent to do it only here though, and it would break BC to change it elsewhere. Deprecate now, alias to `isSuccessful` and remove in Solarium 7? That would be work for a separate PR.

wickedOne:
ugh, those nice little inconsistencies in the solr api...

nevertheless the ``Content-Type`` header seems to be properly set in both cases.

so how about this (and i'm thinking out loud here): when solr responds with anything else than a 200 response code, we override the configured response parser with an ``ErrorResponseParser`` in which we detect the content type which is used to either json_decode / deserialize the response or we use a nifty regex / dom document to extract the necessary information and return it as a generic error object in the response?

this way both scenarios would be covered and can be applied to each solr response.

as for ``getWasSuccessful`` i personally wouldn't mind to introduce a little inconsistency when the intention is to deprecate the call in the first place. alias the method to have both ``getWasSuccessful`` and ``isSuccesful`` and open a ticket to fix that in the future works for me too.

thomascorthals:
The problem I'm seeing right now is that Solarium throws a `HttpException` for status codes >= 400. And that's fine for use cases where you "did something wrong" that resulted in a 4xx.

But for `Exists` specifically, the [Solr ref guide](https://lucene.apache.org/solr/guide/managed-resources.html) states:

> You can test to see if a specific word exists by sending a GET request for that word as a child resource of the set, such as:
> ```bash
> curl "http://localhost:8983/solr/techproducts/schema/analysis/stopwords/english/foo"
> ```
> This request will return a status code of 200 if the child resource (foo) exists or 404 if it does not exist the managed list.

That kind of turns a 404 into a non-error status because it's actually a valid response to your request. Solarium shouldn't throw an exception because the user expects to get just `false` from the library if foo doesn't exist. But if it were a 403 for instance, an exception would still be appropriate because a mere `false` would be very misleading and hide the issue.

I haven't looked into possible failures with other commands yet, and whether they have status codes that would warrant returning `false` without throwing an exception.

I also haven't checked yet what happens exactly with the other queries that implement `getWasSuccessful` (`Server\Collection` and `Server\CoreAdmin`). They don't have any integration tests for failures, only for successes.

mkalkbrenner:
I think we have a bigger issue here - or maybe multiple issues.

According to http://docs.guzzlephp.org/en/stable/quickstart.html#exceptions Guzzle throws an exception for every HTTP code >= 400.

We do the same in the constructor of `Solarium\Core\Query\Result\Result`:
```php
    public function __construct(AbstractQuery $query, Response $response)
    {
        $this->query = $query;
        $this->response = $response;

        // check status for error (range of 400 and 500)
        if ($response->getStatusCode() >= 400) {
            throw new HttpException($response->getStatusMessage(), $response->getStatusCode(), $response->getBody());
        }
    }
```

In our new Psr18Adapter we catch some of guzzle's exceptions and convert them into our own HttpException:
```php
    public function execute(Request $request, Endpoint $endpoint): Response
    {
        try {
            return $this->createResponse($this->httpClient->sendRequest($this->createPsr7Request($request, $endpoint)));
        } catch (ClientExceptionInterface $clientException) {
            throw new HttpException("HTTP request failed, {$clientException}");
        }
    }
```

Here we have an issue that compared to Result::_construct() the response status code is lost.

If we consider guzzle as a reference, it seems that throwing exceptions for 400 and 500 is a common concept.
If you work with REST APIs you need to encapsulate your request in a try-catch-block and examine the response code in the catch.
(BTW that's how we deal with it in Drupal.)

But if we accept this pattern, we still have an issue with the Psr18Adapter.

mkalkbrenner:
https://www.php-fig.org/psr/psr-18/

```
Error handling

A Client MUST NOT treat a well-formed HTTP request or HTTP response as an error condition. For example, response status codes in the 400 and 500 range MUST NOT cause an exception and MUST be returned to the Calling Library as normal.

A Client MUST throw an instance of Psr\Http\Client\ClientExceptionInterface if and only if it is unable to send the HTTP request at all or if the HTTP response could not be parsed into a PSR-7 response object.
```

So guzzle and its behavior isn't a common pattern if we move into the PSR-18 direction in general.

@thomascorthals @wickedOne what's your opinion?

I suggest the keep our current implementation that throws an exception as default, especially for simple searches.
For REST API calls we should overwrite this behaviour and handle (or ignore) the expected response status codes and throw exceptions for the unexpected.

guzzle is not yet PSR-18 compatible, but seems to have an option to turn off the throwing of exceptions. Maybe we should add a hint to our documentation.

wickedOne:
i tend to agree with the psr18 principle that we don't throw any exceptions (other than when there's a network error for example) for well formed responses. if someone uses the guzzle adapter with ``http_errors => true`` i think it's ok to return the exception as a solarium exception with the guzzle exception as previous and let their implementation deal with it.

that would mean deprecation of the constructor logic in ``Solarium\Core\Query\Result\Result`` as @mkalkbrenner pointed out, perhaps something for solarium 7?

in the meantime (i.e. solarium 6.x) we could consider a solr client setting where we can set http_errors to false like guzzle which would bypass throwing the exception in the result class?

@thomascorthals is the fact that the response is either json or html an issue for your usecase? i do have some code i could contribute to normalise that if you're interested

thomascorthals:
@wickedOne For my usecase it doesn't matter what the response format was. I just want to know if the request was successful (returned a 200) or unsuccessful. In case of a 404 for a REST API, unsuccessful doesn't automatically mean an exception is warranted.

thomascorthals:
This now closes #830.

I think I've found an elegant solution for not throwing an exception on 4xx status codes: `Status4xxNoExceptionInterface`. It's an empty marker interface that can be implemented by query classes to signal that they can deal with 4xx responses.

Pros:
- Very easy to add to other queries.
- Behaviour should be obvious from the code.
- Doesn't break BC for existing queries.
- Easily extensible to 5xx statuses.
- Easy to remove if we make "no exceptions" the default behaviour.

Cons:
- Consistency?

I've found a simple solution for the different response format on the managed resource vs. child resource level: don't parse the response body at all if the request wasn't successful. In fact, if the request being successful is all you care about (as with the `Exists` command), a `HEAD` request suffices. Unfortunately, there was a bug introduced in Solr 8.7 where a `HEAD` request returns `HTTP/1.1 200 OK` while the same `GET` returns `HTTP/1.1 404 Not Found`. I've reported this as [SOLR-15116](https://issues.apache.org/jira/browse/SOLR-15116). It's only an issue when querying a child resource so I switch to `GET` for that scenario. I still prefer `HEAD` to check if a managed resource exists because it doesn't download it that way. Some synonym maps I've encountered are several megabytes large.

There was a bug in Solarium that could cause what should have been a `HEAD` request to reach a timeout with '0 of 429 bytes received' when using the cURL adapter. Turns out we didn't use the proper option on the cURL handle for this. I've fixed this along the way. (We did the equivalent of `curl -X HEAD` instead of the equivalent of `curl -I`.)

I've moved `parseStatus()` to `AbstractResponseParser` to avoid duplicate code. It's now available to any response parser that wants to use it.

With the pieces above falling into place, I was able to implement the last bit we were still missing for Managed Resources: requesting a single child resource. Use cases for this include getting a synonym mapping and determining the original case of a term for a case insensitive resource.
```sh
$ curl "http://localhost:8983/solr/techproducts/schema/analysis/synonyms/english/happy"
{
  "responseHeader":{
    "status":0,
    "QTime":2},
  "happy":["glad",
    "joyful"]}

$ curl "http://localhost:8983/solr/techproducts/schema/analysis/stopwords/english/SUCH"
{
  "responseHeader":{
    "status":0,
    "QTime":1},
  "SUCH":"such"}
```
The result from Solr has a different JSON structure than a request for the entire managed resource. As discussed in #830, I've normalised this to use the same result class.

The unit tests have been overhauled and split into actual units. The intergration test no longer has to rely on exceptions to test for nominal behaviour.

mkalkbrenner:
I'm fine with merging this before we deprecate managed resources in the future.

mkalkbrenner (approved):

CHANGELOG.md
```diff
@@ -18,8 +18,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 ## [6.1.3]
+### Added
```
mkalkbrenner, line 21: 6.1.3 is already released ;-)

mkalkbrenner (changes requested):

PR #862, opened 2020-08-16
Feature/proxy support for http adapter

### LINKED ISSUE #861: Missing proxy-Support for Http-Adapter
I noticed that the [Http-Adapter] currently has no proxy support.

This adapter should get proxy implementation similar like it is done for Curl-Adapter.

This can be done by adding this code at [Http-Adapter Line 74]
From: 
```php
        $context = stream_context_create(
            ['http' => [
                    'method' => $method,
                    'timeout' => $this->timeout,
                    'protocol_version' => 1.0,
                    'user_agent' => 'Solarium Http Adapter',
                ],
            ]
        );
```
to
```php
       $http_options = [
              'method' => $method,
              'timeout' => $this->timeout,
              'protocol_version' => 1.0,
              'user_agent' => 'Solarium Http Adapter',
        ];
        if (null !== ($proxy = $this->getOption('proxy'))) {
            $http_options['proxy'] = 'tcp://'.$proxy;
            $http_options['request_fulluri']=>true;
        }
        $context = stream_context_create(['http' =>$http_options]);
```

An example for this change you can see at https://www.php.net/manual/de/function.stream-context-create.php#111032

[Http-Adapter Line 74]: https://github.com/solariumphp/solarium/blob/0e9219daae06fe1da2d2147e23c12a29b46738ec/src/Core/Client/Adapter/Http.php#L74
[Http-Adapter]: https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Adapter/Http.php

Radon8472:
This fixes the missing proxy support for [Http-Adapter].
I updated the documentation ans the examples too.

Unfortumaly it is no longer possible to use `adapteroptions` but you can pass the proxy when creating an adapter:

```php
<?php
// create an HTTP adapter instance
$adapter = new Solarium\Core\Client\Adapter\Http(['proxy'=>'127.0.0.1:8080']);

// ... your requests
```

Closes #861 

[Http-Adapter]: https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Adapter/Http.php


src/Core/Client/Adapter/Http.php
```diff
@@ -12,12 +12,13 @@
...
 use Solarium\Core\Client\Request;
 use Solarium\Core\Client\Response;
+use Solarium\Core\Configurable;
 use Solarium\Exception\HttpException;
 
 /**
  * Basic HTTP adapter using a stream.
  */
-class Http implements AdapterInterface, TimeoutAwareInterface
+class Http extends Configurable implements AdapterInterface, TimeoutAwareInterface
```
wickedOne, line 21: personally i'm not a huge fan of re-introducing the configurable extension here: wouldn't it make more sense to simply use a setter?

src/Core/Client/Adapter/Http.php
```diff
@@ -71,15 +72,26 @@ public function check($data, $headers): void
...
+            $url['scheme'] = 'tcp';
+            // fix behavior when only an IPv4 is passed as proxy, it would be parsed to path not to host
+            if (!isset($url['host']) && isset($url['path'])) {
+                $url['host'] = $url['path'];
+            }
+            if (!isset($url['port'])) {
+                $url['port'] = 8080;
+            }
+            $httpOptions['proxy']           =  sprintf('%s://%s:%d', $url['scheme'], $url['host'], $url['port']);
+            $httpOptions['request_fulluri'] = true;
```
wickedOne, line 92: there's a lot of magic going on here which i think should be taken care of outside this library.
```php
if (null !== $proxy = $this->getProxy()) {
    $httpOptions['proxy'] = $proxy;
    $httpOptions['request_fulluri'] = true;
}
```
should be sufficient

wickedOne: to clarify a bit more: this might work for your specific use case, but for instance makes it impossible to use a http proxy and probably will send those using a transparent proxy on a extended bughunt as the port is hardcoded to 8080 if not set.

the adapter should be responsible for configuring the stream option, not for any business logic surrounding it

src/Core/Client/Adapter/Http.php
Radon8472: I yust see that Curl adapter is using the Configurable, and I think it would be very confusion, when this both adapters dooing the same thing in a different way.

src/Core/Client/Adapter/Http.php
Radon8472: Okay, I aggree.
But what would be the best place to do this Options fixing logic ?
Do you thing it should be part of the configurable ?

Radon8472:
Is anybody checking this PR ?

dmaicher:
I personally would not extend the curl/http adapters any further. We have PSR-18 support and a proxy can easily be configured on a Guzzle Client (that is then used on Solarium via an adapter to PSR-18) for example: https://docs.guzzlephp.org/en/stable/request-options.html#proxy

I think this library should not put too much focus on the http layer implementations as there are various more advanced http client libraries out there doing a good job already :blush: 

@Radon8472 is using a PSR-18 client an issue for you?

Radon8472 (commented):
> 
> 
> I personally would not extend the curl/http adapters any further. We have PSR-18 support and a proxy can easily be configured on a Guzzle Client (that is then used on Solarium via an adapter to PSR-18) for example: https://docs.guzzlephp.org/en/stable/request-options.html#proxy
> 
> I think this library should not put too much focus on the http layer implementations as there are various more advanced http client libraries out there doing a good job already blush
> 
> @Radon8472 is using a PSR-18 client an issue for you?

@dmaicher
In our company we are using solarium in a cms, and it was a very hard piece of work to include solarium in the exusting cms.
So including more libraries is absolut no way, we prefer to use the native php buildin stream functions.

I realy dont understand what is the problem in accepting this PR, it contains successfull tested and perfect working code.

src/Core/Client/Adapter/Http.php
wickedOne: using that sentiment, no library would progress...

the ``ConfigurableInterface`` is somewhat of a god class for this library and though for backwards compatability it might not be possible for us to remove it completely, we should be able to make the decission _not_ to use it

src/Core/Client/Adapter/Http.php
wickedOne: the best place would be when instantiating your adapter. so do your sprintf and port logic outside this library and instantiate your adapter with compiled options

wickedOne (changes requested):
> I realy dont understand what is the problem in accepting this PR, it contains successfull tested and perfect working code.

this is not really helpful; passing tests is good, but not an argument to approve a pull request. 

both myself and @dmaicher tried and provided constructive arguments to make changes of which none are provided...

Radon8472:
> 
> 
> > I realy dont understand what is the problem in accepting this PR, it contains successfull tested and perfect working code.
> 
> this is not really helpful; passing tests is good, but not an argument to approve a pull request.
> 
> both myself and @dmaicher tried and provided constructive arguments to make changes of which none are provided...

This PR adds the missing proxy support, that is already implemented on other adapters, and to add a test, I would need a so setup a proxy that is reachable for everybody.

And @dmaicher did not offer any solution to fix the missing proxy support, he yust suggested not using this adapter.

The facts are simple:
- Your are offering multiple adapters
- this adapter is missing at least one functionality that other adapters already have
- here is a working pull request that adds the missing function

So the big question is, What should be changed on this pull request, that you accept adding the missing functionality to this adapter?

src/Core/Client/Adapter/Http.php
Radon8472: The reason to put this code there was, that the usage of this adapter is similar to the other adapters.
I think it would not be good, that each adapter needs different usage.

wickedOne:
your attidude is really horrible...

the facts are simple:
- this package is free and maintained by various people who are working on it in their leasure time. therefor you're entitled to nothing; the maintainers should agree on the code added to this repository.
- there's nothing stopping you from writing your own adapter implementing the ``AdapterInterface`` and using that in your project
- i've already outlined acceptable changes [on aug 16th](https://github.com/solariumphp/solarium/pull/862#discussion_r471160176)

src/Core/Client/Adapter/Http.php
wickedOne: ?! the only adapter accepting a proxy option inside this library is the curl adapter which takes a single argument and does not apply random magic with protocols and ports. once more, i'm fine with a get / set proxy for the http adapter, but the added "logic" has to go...

mkalkbrenner:
@wickedOne is right about the fact that we should not reintroduce Configurable again.
One major issue with Configurable is that if you switch the adapter for hosting requirements or any other reason, some stuff might silently fail at runtime.

But there's already a pattern we can follow here. Because not every adapter was able to handle client side timeouts we introduced https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Adapter/TimeoutAwareInterface.php and https://github.com/solariumphp/solarium/blob/master/src/Core/Client/Adapter/TimeoutAwareTrait.php

I think we should follow this pattern here and create a ProxyAwareInterface and a ProxyAwareTrait for getter and setter.

wickedOne:
agreed; a proxy aware interface is the way to go on this one. not entirely sure whether a trait is needed for a single argument especially as they need different implementations, but being able to identify an adapter to support proxy usage sounds like a win to me

src/Core/Client/Adapter/Http.php
Radon8472: > 
> ?! the only adapter accepting a proxy option inside this library is the curl adapter which takes a single argument and does not apply random magic with protocols and ports. once more, i'm fine with a get / set proxy for the http adapter, but the added "logic" has to go...

Yes, I am talking about this.
I took the Code from curl adapter as example for this implementation here, that is the reason why I used the `ConfigurableInterface`.
And the stuff with the ports in no "magic", it is common knowledge, that port 8080 is the default port for proxy, if you don pass any custom ports.

Radon8472:
> 
> 
> your attidude is really horrible...
> 
> the facts are simple:
> 
>     * this package is free and maintained by various people who are working on it in their leasure time. therefor you're entitled to nothing; the maintainers should agree on the code added to this repository.
> 
>     * there's nothing stopping you from writing your own adapter implementing the `AdapterInterface` and using that in your project
> 
>     * i've already outlined acceptable changes [on aug 16th](https://github.com/solariumphp/solarium/pull/862#discussion_r471160176)

I know, that everybody could use custom adapters.
But your lib already contains an adapter, that can be used without any extra extensions ect.
The only "problem" is, that at least one function is not correct implemented yet.

I tried to fix this adapter and help other users of this project, to have a full functional adapter that works with native php functions.

wickedOne:
yes you're right and we would welcome the addition of the possibily to set a proxy for the http adapter.
adding a proxy aware interface to both the curl & http adapter would make sense.
things could be as simple as 
```php
<?php

declare(strict_types=1);

/*
 * This file is part of the Solarium package.
 *
 * For the full copyright and license information, please view the COPYING
 * file that was distributed with this source code.
 */

namespace Solarium\Builder\Select;

/**
 * Proxy Aware Interface.
 *
 * @author wicliff <wicliff.wolda@gmail.com>
 */
interface ProxyAwareInterface
{
    /**
     * @param string|null $proxy
     *
     * @return $this
     */
    public function setProxy(?string $proxy): self;

    /**
     * @return string|null
     */
    public function getProxy(): ?string;
}
```
and make both the curl and http adapter use it
```php
// curl adapter
if (null !== $proxy = $this->geProxy()) {
    curl_setopt($handler, CURLOPT_PROXY, $proxy);
}

// http adapter
if (null !== $proxy = $this->getProxy()) {
    $httpOptions['proxy'] = $proxy;
    $httpOptions['request_fulluri'] = true;
}
```

src/Core/Client/Adapter/Http.php
wickedOne: yess, but whenever someone decides to use 8080 for something else, let's say varnish cahe, this would be completely useless. there's no need for this library to decide which proxy port should be used

Radon8472: This lib does not decide wich port is the best, this code yust makes in possible to pass ONLY hostname without port, if this is done, the defaultport is used.
And if you try to connect a proxy on a different port it is possible to add the port.

This is the expected behavior for a proxy setting

dmaicher: > And the stuff with the ports in no "magic", it is common knowledge, that port 8080 is the default port for proxy, if you don pass any custom ports.

not sure I would say its "common knowledge" that port 8080 is a default proxy port. What's your source for that?

AFAIK its just an "alternative port" for http.

See 

https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=109

dmaicher:
> 
> ```php
> // curl adapter
> if (null !== $proxy = $this->geProxy()) {
>     curl_setopt($handler, CURLOPT_PROXY, $proxy);
> }
> ```

@wickedOne but if we want to release this as 6.x without any BC breaks then we need to deprecate the `proxy` option first, or?

src/Core/Client/Adapter/Http.php
Radon8472: The most important refferences should be

https://de.wikipedia.org/wiki/Liste_der_standardisierten_Ports
and 
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=120

There it is listed as http alternative, with the hint that this is mostly used for proxy.

And when you search in google https://www.google.de/search?q=http+proxy+default+port
you get many other sources telling the same.

Radon8472:
> 
> 
> yes you're right and we would welcome the addition of the possibily to set a proxy for the http adapter.
> adding a proxy aware interface to both the curl & http adapter would make sense.
> things could be as simple as
> 
> ```
> <?php
> 
> declare(strict_types=1);
> 
> /*
>  * This file is part of the Solarium package.
>  *
>  * For the full copyright and license information, please view the COPYING
>  * file that was distributed with this source code.
>  */
> 
> namespace Solarium\Core\Client\Adapter;
> 
> /**
>  * Proxy Aware Interface.
>  *
>  * @author wicliff <wicliff.wolda@gmail.com>
>  */
> interface ProxyAwareInterface
> {
>     /**
>      * @param string|null $proxy
>      *
>      * @return $this
>      */
>     public function setProxy(?string $proxy): self;
> 
>     /**
>      * @return string|null
>      */
>     public function getProxy(): ?string;
> }
> ```
> 
> and make both the curl and http adapter use it
> 
> ```
> // curl adapter
> if (null !== $proxy = $this->geProxy()) {
>     curl_setopt($handler, CURLOPT_PROXY, $proxy);
> }
> 
> // http adapter
> if (null !== $proxy = $this->getProxy()) {
>     $httpOptions['proxy'] = $proxy;
>     $httpOptions['request_fulluri'] = true;
> }
> ```

I think thats a good way, but the http adapter needs a bit more code (see https://github.com/solariumphp/solarium/pull/862/files#diff-a1c71d95981bccf3e9ae60e5a0486a758551882998e618f9db14d3f5f4f8ed92R75-R94)
Because the format of the proxy url is different between curl and http stream options.

I would suggest that possible values for the proxy should be at least

- `hostname` (should use http (no ssl) proxy per default on port 8080)
- `hostname:port` (should use (no ssl) http proxy per default the given port)
- `http://hostname:port`
- `https://hostname:port` (is maybe optional, as I read in the current documentation proxy via ssl is currently not ully supported by php)

src/Core/Client/Adapter/Http.php
wickedOne: by your own logic; this library does not decide which is the best and does not set a default for something managed outside this library. so when you don't set a port it will be my.proxy.com and not my.proxy.com:8080

wickedOne:
> > ```
> > // curl adapter
> > if (null !== $proxy = $this->geProxy()) {
> >     curl_setopt($handler, CURLOPT_PROXY, $proxy);
> > }
> > ```
> 
> @wickedOne but if we want to release this as 6.x without any BC breaks then we need to deprecate the `proxy` option first, or?
> 
> Or we make `geProxy()` and `setProxy($proxy)` handle the option internally?

yes the later, or at least i think we should not deprecate the ``proxy`` option for the ``CurlAdapter``. implementing the interface probably could be something like this?
```php
public function getProxy(): string
{
    return $this->options['proxy'];
}

public function setProxy(string $proxy): self
{
    $this->setOption('proxy' $proxy);

    return $this;
}
```

wickedOne:
> I think thats a good way, but the http adapter needs a bit more code (see https://github.com/solariumphp/solarium/pull/862/files#diff-a1c71d95981bccf3e9ae60e5a0486a758551882998e618f9db14d3f5f4f8ed92R75-R94)
> Because the format of the proxy url is different between curl and http stream options.
> 
> I would suggest that possible values for the proxy should be at least
> 
>     * `hostname` (should use http (no ssl) proxy per default on port 8080)
> 
>     * `hostname:port` (should use (no ssl) http proxy per default the given port)
> 
>     * `http://hostname:port`
> 
>     * `https://hostname:port` (is maybe optional, as I read in the current documentation proxy via ssl is currently not ully supported by php)

i don't think we should implement some abstract magic in our code which, based on the adapter someone uses, decides how to create the path for the proxy. userland chooses an adapter and by doing so, should be able to construct the right proxy path and provide it to the adapter.

src/Core/Client/Adapter/Http.php
Radon8472: That is the point. When you enter somewhere "this-address.com" in a http browser, it goes to default port 80,
When you enter it to an ftp software, you get default port 21, when you enter it in vnc you get default port 5900.

So in every software, a defaultport is choosen automatically, when you dont specify a port.
So why it should not be done in this lib ?

src/Core/Client/Adapter/Http.php
Radon8472: And what is you suggestion not, to finaly finish this pull-request?

PR #869, opened 2020-09-05
Fix return typehint for ClusterState::getCollections()

tsterker:
In case there are no collections yet, a `TypeError` is thrown:
```
TypeError: Return value of Solarium\Core\Client\State\ClusterState::getCollections() must be of the type array, null returned
```

### DIFF
--- a/src/Core/Client/State/ClusterState.php
+++ b/src/Core/Client/State/ClusterState.php
@@ -121,7 +121,7 @@ class ClusterState
     }
 
     /**
-     * @return ClusterState[]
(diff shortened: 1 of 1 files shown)


tsterker:
I just realized that the `$collections` property is not initialized to an empty array and thus `getCollections()` could return `null`.

I updated the PR accordingly, but am wondering if it would make sense to just initialize `$collections` to an empty array?

mkalkbrenner:
Thanks for the fix. But you're right that the better fix is to initialize with an empty array. See #875

Issue #870, opened 2020-09-07
Missing functionality for nested documents associated via a field "comment" .


davchs:
**Solarium version(s) affected**: 6.0.1
**Solr version**:8.5.2

**Description**  
Depending on the Solr Documentation [Indexing Nested Child Documents](https://lucene.apache.org/solr/guide/8_6/indexing-nested-documents.html#schema-configuration)  there are two ways to create a nested document but solarium only handle one. 

> It illustrates two styles of adding child documents: the first is associated via a field "comment" (preferred), and the second is done in the classic way now referred to as an "anonymous" or "unlabelled" child document.

Solarium would create a xml like this :
```
<add>
   <doc>
      <field></field>
      <field></field>
      <doc>
      </doc>
      <doc>
      </doc>
   </doc>
</add>
```
this is correct for the second solution 
But solarium doesn't handle the solution "associated via a field "comment" (preferred)" where you would need a structure like this: 
```
<add>
   <doc>
      <field></field>
      <field></field>
      <field>
          <doc>
          </doc>
      </field>
      <field>
           <doc>
           </doc>
      </field>
   </doc>
</add>
```

I'm currently working in a project where I ran into problems with this. In my project i patched the [RequestBuilder](https://github.com/davchs/solarium/commit/6230b4c3e3bbcbcad27625a8172297e80dd63709) to handle it that way. But this would then only handle nested documents associated via a field "comment". I try to find another possible Solution how Solarium could handle both. need tbd

thomascorthals:
It just so happens that I added an integration test for nested documents last month using the preferred style.

https://github.com/solariumphp/solarium/blob/4bdd5aa674641a0751ba7a5846c5b9e31f5c70c9/tests/Integration/AbstractTechproductsTest.php#L958-L996

Can you try and modify this code to work with your data? The test passes against Solr 7 and 8, but maybe I'm not testing (everything) what I'm supposed to be testing.

davchs:
The `$data` in your Snippet would lead to the following XML in Solarium (anonymous children) :
```
<add>
<doc>
    <field name="id">solarium-parent</field>
    <field name="name">Solarium Nested Document Parent</field>
    <field name="cat">solarium-nested-document</field>
    <field name="cat">parent</field>
    <doc>
        <field name="id">solarium-child-1</field>
        <field name="name">Solarium Nested Document Child 1</field>
        <field name="cat">solarium-nested-document</field>
        <field name="cat">child</field>
        <doc>
[...]

PR #872, opened 2020-09-08
 Add functionality for nested documents associated via a field "comment" 

davchs:
As mentioned in https://github.com/solariumphp/solarium/issues/870 I opened a PR to add functionality for nested documents. 
To support a document structure like this: 
```
<add>
  <doc>
    <field name="ID">1</field>
    <field name="title">Solr adds block join support</field>
    <field name="content_type">parentDocument</field>
    <field name="content">
      <doc>
        <field name="ID">2</field>
        <field name="comments">SolrCloud supports it too!</field>
      </doc>
    </field>
  </doc>
</add>
```
unlabelled children could be added with the special field key `_childDocuments_` like this: 
```
 $update = $client->createUpdate();
        $doc1 = $update->createDocument();
        $doc1->id = uniqueid();
        $doc1->type = 'parent';
        $doc1->title = 'I am parent';
        $doc1->_childDocuments_ = [
            [
                'id' => uniqueid(),
                'type' => 'child',
                'title' => 'I am child 1',
            ],
            [
                'id' => uniqueid(),
                'type' => 'child',
                'title' => 'I am child 2',
            ],
        ];
```
labelled children could be added like this: 
```
 $update = $client->createUpdate();
        $doc1 = $update->createDocument();
        $doc1->id = uniqueid();
        $doc1->type = 'parent';
        $doc1->title = 'I am parent';
        $doc1->childs = [
            [
                'id' => uniqueid(),
                'type' => 'child',
                'title' => 'I am child 1',
            ],
            [
                'id' => uniqueid(),
                'type' => 'child',
                'title' => 'I am child 2',
            ],
        ];
```


tbolier:
@mkalkbrenner Can you please check and merge this pull request? Right now solarium is not compatible with named (labelled) nested documents of SOLR 8, making it impossible to index documents in SOLR 8 with the new nested child documents in a path.

For more information see: https://lucene.apache.org/solr/guide/8_6/indexing-nested-documents.html

tbolier:
@davchs I have added a patch here https://github.com/davchs/solarium/pull/1 for backwards compatibility for anonymous child documents. Please have a look, once you merge it, it will be added to this PR.

thomascorthals:
I'd like to weigh in on @tbolier's PR. The documentation for nested documents has received an overhaul for Solr 8.7 and it's much clearer now. I've compared the ref guide for 7, 8.6 and 8.7 when working on #889 and had a peek at the source code. The only way to add anonymous children through JSON is with a `_childDocuments_` pseudo-field. It's hardcoded in the JsonLoader and commented for Solr 8 as `// somewhat legacy`. There is no special treatment for this field name in the XMLLoader.

In my opinion, it's a valid choice to mirror this in Solarium by reserving `_childDocuments_` for anonymous children. It's in line with the Sorl ref guide for JSON requests, and being a library our users don't need to concern themselves with whether we use JSON or XML as long as our documented behaviour is predictable and consistent. If you really must add them labelled like this, you could still do it like this:
```php
$update->addRawXmlCommand('<add><doc><field name="id">solarium-parent</field><field name="_childDocuments_"><doc><field name="id">solarium-child-1</field></doc><doc><field name="id">solarium-child-2</field></doc></field></doc></add>');
```

The opposite is also true. We could assume the user did intend it as an actual label and then a raw XML command could be used to add anonymous children. I think this would be an annoyance for more users than the other way around. Based on the ref guide, and oblivious to Solarium (or Solr for that matter) internals, one would be more likely to assume using `_childDocuments_` results in anonymous children. The revamped ref guide for Solr 8.7 emphasises the JSON examples, if you don't switch to XML explicitly you'll never know there's a difference. There is nothing in the ref guide that confirms it will be used as an actual label with an XML request. I honestly don't think a novice user would attempt to use it that way based on that information.

@mkalkbrenner If we keep the functionality for anonymous alongside labelled children, #889 should be merged first as it will prove backward compatibility for this PR. If this PR gets merged, I'll be happy to provide the integration test for the full CRUD cycle including atomic updates ASAP.

mkalkbrenner:
> If we keep the functionality for anonymous alongside labelled children, #889 should be merged first as it will prove backward compatibility for this PR. If this PR gets merged, I'll be happy to provide the integration test for the full CRUD cycle including atomic updates ASAP.

@thomascorthals here we go ...

mkalkbrenner:
@thomascorthals @davchs @tbolier
As requested I merged #889. And I think that you agreed on the implementation. As soon as it is finished I'll merge it.

thomascorthals:
@mkalkbrenner It would be great if you could merge it without the integration test. I can't fork @davchs's repo because it is itself a fork of a repo I already forked. That would allow me to open a PR for it without having to work around that limitation of GitHub.

tbolier:
@davchs @thomascorthals @mkalkbrenner 

Thank you all, I was a little bit busy with other projects, so have been out of the loop. I'm happy to read the solution got accepted and the code was merged. Great! Happy holidays!

PR #879, opened 2020-10-23
Use boolean instead of int as second parameter to print_r()

### LINKED ISSUE #878: [BUG] print_r() expects parameter 2 to be bool, int given
**Solarium version(s) affected**: all with `print_r()` usage
**Solr version**: not relevant

**Description**  
PHP throws exception

**How to reproduce**  
Execute the code with `print_r('some output', 1)` usages.

**Possible Solution**  
Change all occurrences with `print_r('some output', 1)` to `print_r('some output', true)` 

**Additional context**  
See https://github.com/TYPO3-Solr/ext-solr/issues/2753

https://github.com/solariumphp/solarium/blob/5b2cc069b30bd990e9130e7a3b8c54ce4b1fe04a/src/Core/Client/Request.php#L91

and

https://github.com/solariumphp/solarium/blob/5b2cc069b30bd990e9130e7a3b8c54ce4b1fe04a/src/Core/Client/Endpoint.php#L48

thomascorthals:
Closes #878.

### DIFF
--- a/src/Core/Client/Endpoint.php
+++ b/src/Core/Client/Endpoint.php
@@ -45,7 +45,7 @@ class Endpoint extends Configurable
      */
     public function __toString()
     {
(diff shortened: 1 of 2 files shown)


thomascorthals:
The error only gets thrown with `declare(strict_types = 1);` by the way.

PR #880, opened 2020-10-30
allow php 8

dmaicher:
This makes solarium PHP 8 compatible and allows to use it with PHP 8.

**Note:** there are some code changes that might be considered BC breaks. So maybe we should discuss this.

TODOs:
- [x] run CI tests with php 8 / nightly
- [x] we should first merge https://github.com/solariumphp/solarium/pull/881

### DIFF
--- a/src/Core/Query/AbstractDocument.php
+++ b/src/Core/Query/AbstractDocument.php
@@ -24,10 +24,8 @@ abstract class AbstractDocument implements DocumentInterface, \IteratorAggregate
     /**
      * @param mixed $name
      * @param mixed $value
-     *
-     * @return \Solarium\Core\Query\DocumentInterface
      */
-    abstract public function __set($name, $value): DocumentInterface;
+    abstract public function __set($name, $value): void;
 
     /**
      * Get field value by name.
--- a/src/QueryType/Update/Query/Document.php
+++ b/src/QueryType/Update/Query/Document.php
@@ -175,14 +175,10 @@ class Document extends AbstractDocument
      *
      * @param string $name
      * @param mixed  $value
-     *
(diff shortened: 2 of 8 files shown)


src/Core/Query/AbstractDocument.php
```diff
@@ -24,10 +24,8 @@ abstract class AbstractDocument implements DocumentInterface, \IteratorAggregate
     /**
      * @param mixed $name
      * @param mixed $value
-     *
-     * @return \Solarium\Core\Query\DocumentInterface
      */
-    abstract public function __set($name, $value): DocumentInterface;
+    abstract public function __set($name, $value): void;
```
dmaicher, line 28: Can this be a BC break. Not sure if anyone can be relying on the return type here. Maybe yes.

But it definitely was invalid even on php7 and should have never worked like this.

See https://www.php.net/manual/en/language.oop5.overloading.php#object.set



WDYT?

src/QueryType/Update/Query/Document.php
```diff
@@ -191,14 +186,10 @@ public function __set($name, $value): DocumentInterface
      * Magic method for removing fields by un-setting object properties
      *
      * @param string $name
-     *
-     * @return self
      */
-    public function __unset($name): self
+    public function __unset($name): void
```
dmaicher, line 190: See https://www.php.net/manual/en/language.oop5.overloading.php#object.unset

src/Core/Query/AbstractDocument.php
mkalkbrenner: According to the documentation this is rather a bug fix then a BC break.

Issue #397, opened 2016-02-16
reuse the curl connection handlers


domnulnopcea:
Hi,

If we do multiple solr queries (tens of thousands) that means opening and closing thousands of curl connections.

this is bad!

why don't reuse the same curl handle!

do you plan to do this?

thanks

basdenooijer:
Well, most use cases only have one Solr query in an application request, so maybe that's why you are the first one to run into (or to report) this issue.

I agree in some use cases it would be useful to reuse the Curl handle, so this could be made configurable. 

A pull request with this functionality would be very welcome

domnulnopcea:
hm...I can not see a case where you would not want to reuse the curl handlers. I do not see a point where we want this to be configurable. You just end up eating less resources by reusing the handlers...

can I send a patch where this is the default behaviour without any other configuration addons?

thanks

basdenooijer:
That case would be when you never intend to re-use it, for example when you do only one query and continue a long running script that never needs to re-query Solr. If the connection is not closed you would leave a socket open for some time, where in some cases it would be better to close resources as soon as you're done.
Since the library cannot know in advance if you are done with Solr queries, the only safe option is to close the the connection by default. 

I think the scenario you describe (with reuse being better) is actually more common, but the library cannot change behaviour except for a major release. Therefore I suggest a configuration option with a default value 'off'. Maybe in the next major release this option could be changed to 'on' by default.

MaxamusOne:
I have used connection reusing a while ago. Let me come back to you with the solution.
The use case was reindexing a quite big full-text index and sending thousands of update queries to solr.

domnulnopcea:
great! thanks a lot

domnulnopcea:
any news here?

thomascorthals:
> any news here?

@domnulnopcea Solarium 6.0 will deprecate the Curl adapter in favour of a [PSR-18](https://www.php-fig.org/psr/psr-18/) compatible implementation as described in #550. You can follow the implementation progress in PR #754.

dmaicher:
@thomascorthals @domnulnopcea we actually did not deprecate the `Curl` adapter in 5.x and it will still exist in 6.0.

But I would argue that we should not further extend its functionality now that we have PSR-18 support. Re-using an existing connection can probably be easily achieved by using some properly configured PSR-18 client and the now existing `Psr18Adapter`.

thomascorthals:
[...]

PR #887, opened 2020-11-17
Extend documentation for PSR-18 adapter: HTTP connection reuse and custom timeouts

tsterker:
In reference to #397.

Given the performance improvements I had, I think this topic could be addressed more prominently. But such a tiny note could already go a long way :)


docs/client-and-adapters.md
```diff
@@ -108,6 +108,8 @@ htmlFooter();
 
 ```
 
+See [#397](https://github.com/solariumphp/solarium/issues/397#issuecomment-727662811) on how to leverage the [Symphony PSR-18 HTTP Client](https://symfony.com/doc/current/http_client.html#psr-18-and-psr-17) to reuse Solr connections.
```
wickedOne, line 111: you might want to give the example from your comment here rather than referencing it?

tsterker: @wickedOne Sorry for the silence. I went ahead with a proposal to extend the documentation for the PSR-18 Adapter, addressing:

- Custom timeouts
- Connection reuse
- Code additional example
- Note on open file handles "leak"

Feedback/suggestions welcome.

PR #896, opened 2020-12-13
Auto-detect point values and remove $escape parameter in…

### LINKED ISSUE #895: Escaping range queries can cause confusion
I ran into this issue with result grouping. I can easily code around it as a user, but I think we should fix it as a library.

```php
$grouping = $select->getGrouping();
$grouping->addQueries([
    $select->getHelper()->rangeQuery('price', 0, 99.99),
    $select->getHelper()->rangeQuery('price', 100, null),
]);

$result = self::$client->select($select);
$groupingResult = $result->getComponent(ComponentAwareQueryInterface::COMPONENT_GROUPING);

/** @var QueryGroup $queryGroup */
$queryGroup = $groupingComponentResult->getGroup('price:[0 TO 99.99]');
```

Based on the [Solr ref guide](https://lucene.apache.org/solr/guide/result-grouping.html#grouping-by-query), you would assume this gets you the first query group. But you actually get `null` because the query that was sent to Solr is `price:["0" TO "99.99"]`.

https://github.com/solariumphp/solarium/blob/58211835260e6aa1c3d180b4dc3ca8567290de06/src/Core/Query/Helper.php#L180-L182

I get why this is necessary for strings. For integers and floats however, there are no security implications that I can think of when bypassing the escaping. In fact, it might be more secure to abandon this parameter altogether. If point values aren't properly validated from user input, the security risk isn't mitigated if we explicitly tell to disable it. We could detect a point value and handle escaping transparently without the extra parameter.

In addition, there's an error in the PHPdoc.

https://github.com/solariumphp/solarium/blob/58211835260e6aa1c3d180b4dc3ca8567290de06/src/Core/Query/Helper.php#L173-L174

That's actually `store:{"5" TO "*"}`. (And it should be.)

My proposal:
- `null` remains `*`
- integers and floats are used unescaped
- point values are detected by regex `/^-?\d+(?:\.\d+)?[, ]-?\d+(?:\.\d+)?$/` and unescaped
- everything else is escaped

Am I missing something here?

thomascorthals:
… $helper->rangeQuery()

I've implemented my proposal to close #895.

Removing the `$escape` parameter doesn't break BC because it's just ignored when still passed. Or would it be better to throw a deprecation warning when it's passed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Exception message for invalid/unavailable file in Extract query now contains filename
+- Helper::rangeQuery() detects point values without parameter to turn off escaping
 
 ### Removed
 
--- a/docs/queries/query-helper/query-helper.md
+++ b/docs/queries/query-helper/query-helper.md
@@ -7,7 +7,7 @@ Two special types of helper methods are
 Helper methods for general use
 ------------------------------
 
--   rangeQuery($field, $from, $to, $inclusive = true, $escape = true)
+-   rangeQuery($field, $from, $to, $inclusive = true)
 -   qparser($name, $params = array())
 -   functionCall($name, $params = array())
 -   join($from, $to, $dereferenced = false)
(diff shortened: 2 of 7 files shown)


mkalkbrenner:
If I remember correctly there was a security issue if you don't "escape" the values. But the validation in the PR looks good enough.

PR #900, opened 2020-12-23
Left-inclusive only and right-inclusive only range queries

thomascorthals:
Solr allows range queries to be left-inclusive or right-inclusive (e.g. `price:{0 TO 10]`, `price:[0 TO 10}`). Adding an additional parameter to `rangeQuery()` wouldn't be backward compatible because there used to be one with a different meaning. Even if that wasn't the case, it couldn't have a straightforward default value in a BC manner.

The best way I could come up with is an array with two booleans for left and right while still allowing a single boolean as shorthand for `[true, true]` or `[false, false]`.


mkalkbrenner:
@thomascorthals thanks a lot!

Issue #610, opened 2018-08-01
PrefetchIterator Behavior Explanation 


CBox:
Hey,

I wonder how the Prefetch plugin works, it will iterate everything and get the results back or while we iterate it sends the request?

The reason why it's make a difference is if I pull 1M records all of them going to be saved in the "documents" variable or only the last fetch? (more data to memory)

What's going on when I do count($prefetch)? It's iterate all the result or just return the numFound?

Also is this plugin can work with ParallelExecution plugin?

Thanks a lot!

mkalkbrenner:
The prefetch plugin uses cursormark and should not fetch all documents. But since the latest release you can also leverage cursormark on standard select queries without using the prefetch plugin.

CBox:
It means that I will be able to process all the 1M records without save them to a variable but it will save the last fetch?

CBox:
Last question is the ParallelExecution can work with the PrefetchIterator plugin?

mkalkbrenner:
I assume yes. But I never tried it.

mkalkbrenner:
I recommend to use the standard select with cursormark. That's newer code ;-)

CBox:
Is there any sample code I can take a look at?
Thank you again :)

mkalkbrenner:
We have an example that uses the combination:
https://github.com/solariumphp/solarium/blob/master/examples/7.6-plugin-prefetchiterator.php

I suggest you simply give it a try. And it would be nice if you report back if it works like expected.

CBox:
Just to make sure, if I'm not using this:
`$query->setCursormark('*');
// cursor functionality requires a sort containing a uniqueKey field as tie breaker on top of your desired sorts for the query
$query->addSort('id', $query::SORT_ASC);`

The Prefetch plugin not adding it? I was sure that it's adding it automatically, I've used the sort but not the setCursormark - it mean it was not active right?

CBox:
I guess I missed it here:
`if (null === $this->cursormark) {
            $this->query->setStart($this->start)->setRows($this->getPrefetch());
        } else {
            $this->query->setCursormark($this->cursormark)->setRows($this->getPrefetch());
        }`

Still if you can please verify :)

CBox:
After adding $query->setCursormark('*'); I didn't not get duplicates but there was a lot of data missing.

Can I use my own pagination instead of Prefetch? I will add and update the setCursormark, is it possible?

Thanks!

thomascorthals:
[...]

PR #902, opened 2020-12-27
Bugfix for uninitialised position in PrefetchIterator

thomascorthals:
`PrefetchIterator` didn't have an initial value for its `$position` and that causes two bugs. You can test it by replacing the `foreach` loop in `examples/7.6-plugin-prefetchiterator.php`.

1) You got an error when calling `key()` on a fresh iterator if you didn't call `rewind()` first.

```php
for (; $prefetch->valid(); $prefetch->next()) {
    echo '<hr/>'.$prefetch->key().': '. $prefetch->current()->id;
}
```
```
Fatal error: Uncaught TypeError: Return value of Solarium\Plugin\PrefetchIterator::key() must be of the type int, null returned in solarium/src/Plugin/PrefetchIterator.php:205 Stack trace: #0 solarium/examples/7.6-plugin-prefetchiterator.php(26): Solarium\Plugin\PrefetchIterator->key() #1 {main} thrown in solarium/src/Plugin/PrefetchIterator.php on line 205
```

2) If you're doing a `count()` first and then loop through it without `rewind()`ing, the first set of fetched documents would be skipped. In our example, `$i` ended up being 30 for the 32 documents in techproducts.

```php
for ($i = 0; $prefetch->valid(); $prefetch->next(), ++$i) {
    echo '<hr/>'.$i.': '. $prefetch->current()->id;
}
```

These problems don't surface when you're using `foreach` to loop over the results because that does a `rewind()` first. But you would run into it with a `while` in a use case like this:
```php
while ($n < $max && $prefetch->valid()) {
    if (doCheck($prefetch->current()) {
        ++$n;
    }
    $prefetch->next();
}
```

Furthermore, since `count()` implements a method from `\Countable`, it shouldn't have any publically visible effects on the methods that are implemented from `\Iterator`. They're separate interfaces.

Both problems disappear when `$position` is initialised to `0`. In fact the [example in the PHP Manual](https://www.php.net/manual/en/class.iterator.php#iterator.examples) does just that.

I've added a manual walkthrough of a full iteration and some rewinds in the unit tests. I've also added tests to ensure an empty result set works as expected.

I do believe I must've run into the same issue when I first added the integration tests. I didn't include an `assertCount()` then and I think it was because it didn't agree with the loop. It's included now and the loop also checks if the documents are actually iterated in the correct order. This is a relevant test for #610.

### DIFF
--- a/src/Plugin/PrefetchIterator.php
+++ b/src/Plugin/PrefetchIterator.php
@@ -58,7 +58,7 @@ class PrefetchIterator extends AbstractPlugin implements \Iterator, \Countable
      *
      * @var int
      */
-    protected $position;
+    protected $position = 0;
 
     /**
      * Cursor mark.
@@ -262,7 +262,7 @@ class PrefetchIterator extends AbstractPlugin implements \Iterator, \Countable
      */
     protected function resetData(): self
     {
-        $this->position = null;
+        $this->position = 0;
         $this->result = null;
         $this->documents = null;
         $this->start = 0;
--- a/src/Plugin/PrefetchIterator.php
+++ b/src/Plugin/PrefetchIterator.php
@@ -173,7 +173,7 @@ class PrefetchIterator extends AbstractPlugin implements \Iterator, \Countable
     {
         $this->position = 0;
 
-        // this condition prevent useless re-fetching of data if a count is done before the iterator is used
+        // this condition prevents erroneously fetching the next set of results if a count is done before the iterator is used
         if ($this->start !== $this->options['prefetch']) {
             $this->start = 0;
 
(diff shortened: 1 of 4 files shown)


src/Plugin/PrefetchIterator.php
```diff
@@ -58,7 +58,7 @@ class PrefetchIterator extends AbstractPlugin implements \Iterator, \Countable
      *
      * @var int
      */
-    protected $position;
+    protected $position = 0;
```
wickedOne, line 61: isn't defining the default in the constructor sufficient?

thomascorthals: The example in the PHP Manual does both:
```php
<?php
class myIterator implements Iterator {
    private $position = 0;
    // ...  

    public function __construct() {
        $this->position = 0;
    }
```

As it is now, none of the plugin classes have an explicit constructor defined. The parent constructor — `Configurable::__construct()` — is called from `AbstractPlugin::__construct()`.

Following the logic of `Configurable`, this should be done in `PrefetchIterator::init()`? But then we might as well rename `resetData()` to `init()` …

wickedOne: ah yes, i see...

personally i think ``resetData`` is fine

wickedOne (approved):

PR #903, opened 2020-12-28
drop support for php 7.2

wickedOne:
php 7.2 eol was now 30th
https://www.php.net/eol.php

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -14,7 +14,7 @@ jobs:
 
         strategy:
             matrix:
-                php: [7.2, 7.3, 7.4, 8.0]
+                php: [7.3, 7.4, 8.0]
                 solr: [7, 8]
                 mode: [cloud, server]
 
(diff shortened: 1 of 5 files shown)


thomascorthals:
You'll have to bump the minimal required version in `README.md` and `docs/getting-started.md` as well.

thomascorthals:
Actually, since it's worded like this …
https://github.com/solariumphp/solarium/blob/7cf99c98523e861e1e43ed7e58309b2763a144a9/README.md#L13

Should this be considered a BC break for Solarium 6.x?

wickedOne:
personally i think the minor should increment to 6.1.0.

semver.org is not really clear about it;
>Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API.

and as we don't change our api, just the requirements (of which there's no explicit mention) i'm not sure.

one of the reasons why i'm willing to bump to 7.3 is to be able to use the ``JSON_THROW_ON_ERROR`` flag in the json encode function without builds failing.

dmaicher (approved):

mkalkbrenner:
I think we should go with 6.1.0.
BTW I know that we still have solarium 3 users because of the supported PHP version.

wickedOne:
ugh, i do not envy those users...

PR #911, opened 2021-01-08
Locale-independent treatment of floats

thomascorthals:
This is a proof-of-concept to discuss in #909, not a finished implementation.


thomascorthals:
With #923 merged, this one's no longer relevant.

PR #914, opened 2021-01-13
Fix for "maxScore":"NaN" (SOLR-13839)

### LINKED ISSUE #913: When a grouping group returns 0 results solarium raise exception
**Solarium version(s) affected**: 6.0.4
**Solr version**:8.6.3

**Description**  
I get this error:
`
 Severity: error --> Exception: Argument 4 passed to Solarium\Component\Result\Grouping\QueryGroup::__construct() must be of the type float or null, string given, called in /Users/alendo/Sites/didatticaweb/vendor/solarium/solarium/src/Component/ResponseParser/Grouping.php on line 102 /Users/alendo/Sites/didatticaweb/vendor/solarium/solarium/src/Component/Result/Grouping/QueryGroup.php 71
`
**How to reproduce**  
The query result has a group with 0 documents, like this:
``
{
  "grouped":{
    "tipo_s:corso":{
      "matches":66,
      "doclist":{"numFound":0,"start":0,"maxScore":"NaN","numFoundExact":true,"docs":[]
      }}...
``
**Possible Solution**  
The error is raised because $maxScore value is "NaN".
I added this row before 
/Users/alendo/Sites/didatticaweb/vendor/solarium/solarium/src/Component/ResponseParser/Grouping.php line 102:

if (is_string($maxScore)) $maxScore=null;

**Additional context**

thomascorthals:
Solr returns `"maxScore":"NaN"` when `group.query` doesn't match any docs in standalone/single shard mode with `score` in `fl` ([SOLR-13839](https://issues.apache.org/jira/browse/SOLR-13839)). I've caught this in the response parser to turn it into `null` as the Solarium methods working with this value expect a `float` or `null`. (Actually, I had to fix one of those methods to handle `null` as well.)

I've added unit and integration tests, extended existing tests to cover `maxScore` (and `start` while I was at it), placed comments to reference SOLR-13839 in the source code, and added a note in the `CHANGELOG`. This closes #913.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - PrefetchIterator::key() should return 0 instead of NULL on a fresh PrefetchIterator
 - PrefetchIterator::next() shouldn't skip fetched results after PrefetchIterator::count() on a fresh PrefetchIterator
 - PrefetchIterator::rewind() no longer results in duplicate documents when invoked mid-set
-- fixed incorrect median function
+- Fixed incorrect median function
+- Fix for maxScore being returned as "NaN" when group.query doesn't match any docs (SOLR-13839)
 
 ### Changed
 - Exception message for invalid/unavailable file in Extract query now contains filename
--- a/src/Component/ResponseParser/Grouping.php
+++ b/src/Component/ResponseParser/Grouping.php
@@ -89,6 +89,14 @@ class Grouping implements ComponentParserInterface
                 $start = $result['doclist']['start'] ?? null;
                 $maxScore = $result['doclist']['maxScore'] ?? null;
 
+                /*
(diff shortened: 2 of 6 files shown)


thomascorthals:
Oh great, when the issue said
> - This happens only in standalone/single shard mode
I assumed it would be `null` or not returned at all for SolrCloud. Apparently it's `1.0`, but I'm reluctant to assume that in the test because I couldn't find this documented.

@mkalkbrenner Would it be better to skip this test for SolrCloud?

thomascorthals:
I went ahead and added `@group skip_for_solr_cloud`. I've documented why, with a link to SOLR-6612. Also watching SOLR-6612 on their JIRA.

Whenever one of my watched issues gets fixed in Solr (hasn't happened to me yet), I will search for the issue number in Solarium. That should jog my memory and allow me to re-evaluate these exceptions in the tests.

PR #922, opened 2021-02-27
allow Local Parameters like countDistinct

Raggok:
with the actual Result we cant get local parameters like countDistinct, there is no need to keep the function as protected.

$stats = $query->getStats();
$stats->createField($field)->addPivot('tag countDistinct=true');

that parse into

stats.field={!tag=tag countDistinct=true}field

and actually there is no way to get that stats value

### DIFF
--- a/src/Component/Result/Stats/Result.php
+++ b/src/Component/Result/Stats/Result.php
@@ -157,7 +157,7 @@ class Result
      *
      * @return string|array|null
      */
-    protected function getValue($name)
+    public function getValue($name)
     {
         return $this->stats[$name] ?? null;
     }


mkalkbrenner (changes requested):
This look good. But we should have an additional test for this.

thomascorthals:
@Raggok [Solarium 6.1.5](https://github.com/solariumphp/solarium/releases/tag/6.1.5) was released with improved support for the Stats Component. There's a `getCountDistinct()` method that you can use to get the stats value for the `countDistinct` parameter.

Your solution to make `getValue()` a public method is still valid. It ensures that any new stats values that might be added in future Solr releases will be accessible even if they don't have a dedicated method in Solarium yet. But I had to rename it to `getStatsValue()` (and deprecate `getValue()`) to avoid a name conflict in code that is shared between regular stats results and field results for pivot faceting.

Issue #713, opened 2019-10-30
Pivot Facet Limit does not work


cseufert:
I have updated to Solarium 5.1.4 to take advantage of the api for setting limits on Pivot Facet, but it still doesn't work.

I am setting up a PivotFacet & a regular Facet like this:
```
    $facetObs = $facet->createFacetPivot('question-obs');
    $facetObs->addFields(['qcode', 'observation']);
    $facetObs->setLimit(-1);
    $facetObs->setMinCount(0);
  $facet
      ->createFacetField('report')
      ->setField('report')
      ->setMinCount(1)
      ->setLimit(-1);
```
however, it never gets sent to the server, eg:
`2019-10-30 05:56:09.598 INFO  (qtp1330278544-33) [   x:inspect] o.a.s.c.S.Request [inspect]  webapp=/solr path=/select params={f.report.facet.mincount=1&facet.field={!key%3Dreport}report&json.nl=flat&facet.pivot={!key%3Dquestion-obs}qcode,observation&fl=*,score&start=0&rows=10&facet.pivot.mincount=0&q=(inspection:110f2b70eec711e8baa3bc305bf50a50)+AND+(vessel:9294109)+AND+reportdate:+[2019-10-24T00:00:00Z+TO+2019-10-26T23:59:59Z]&omitHeader=true&wt=json&facet=true&f.report.facet.limit=-1} hits=243 status=0 QTime=1`

As a side note, I can't figure out why the parameter is not being sent to the server, I cant see how the facet parameters are turned into query parameters, if you can point me in the right direction, I could possibly fix it.

wickedOne:
fixed @ https://github.com/solariumphp/solarium/pull/732

PR #924, opened 2021-04-01
(WIP) fix: Facet pivots don't support limits

tsterker:
**Update** (2021-04-04): I realized this feature was explicitly added in https://github.com/solariumphp/solarium/pull/732 and pointed out as missing in https://github.com/solariumphp/solarium/issues/713.
Now I wonder if I am missing something, but for me the [Solr documentation](https://solr.apache.org/guide/8_6/faceting.html#additional-pivot-parameters) seemed rather clear and also [google](https://www.google.com/search?q=%22facet.pivot.limit%22) does not really yield anything for `facet.pivot.limit`.

---

> ⚠️  This PR is **work-in-progress**:
> - Did not run/update any tests
> - Not sure if I might be missing something in the Solr specification

I just hunted down a bug in my app where I did **not properly set the facet limits**.

Calling `setLimit` on a `Pivot` results in a `facet.pivot.limit` parameter to be set in the Solr request, which does not seem to be a supported parameter. Instead, the limit must be set on the `FacetSet` itself (which will yield `facet.limit`). See [Solr documentation](https://solr.apache.org/guide/8_6/faceting.html#additional-pivot-parameters) with a reference to this.

Unfortunately, I currently can't spend more time on this topic (investigation/implementation), but wanted to already start a conversation with this PR.

### DIFF
--- a/src/Component/Facet/Pivot.php
+++ b/src/Component/Facet/Pivot.php
@@ -37,28 +37,6 @@ class Pivot extends AbstractFacet
     }
 
     /**
-     * Set the facet limit.
-     *
-     * @param int $limit
-     *
-     * @return self Provides fluent interface
-     */
-    public function setLimit($limit): self
-    {
-        return $this->setOption('limit', $limit);
-    }
-
-    /**
-     * Get the facet limit.
-     *
-     * @return int
-     */
-    public function getLimit(): ?int
-    {
-        return $this->getOption('limit');
-    }
-
-    /**
      * Set the facet mincount.
      *
      * @param int $minCount
--- a/src/Component/RequestBuilder/FacetSet.php
+++ b/src/Component/RequestBuilder/FacetSet.php
@@ -267,7 +267,6 @@ class FacetSet extends RequestBuilder implements ComponentRequestBuilderInterfac
             sprintf('%s%s', $facet->getLocalParameters()->render(), implode(',', $facet->getFields()))
         );
         $request->addParam('facet.pivot.mincount', $facet->getMinCount(), true);
-        $request->addParam('facet.pivot.limit', $facet->getLimit(), true);
     }
 
     /**


thomascorthals:
> ### Additional Pivot Parameters
> 
> Although `facet.pivot.mincount` deviates in name from the `facet.mincount` parameter used by field faceting, many of the faceting parameters described above can also be used with pivot faceting:
> 
> - `facet.limit`
> - `facet.offset`
> - `facet.sort`
> - `facet.overrequest.count`
> - `facet.overrequest.ratio` 

This does seem to say that `facet.pivot.mincount` is the only parameter with a  `.pivot` tmesis. It could be written a bit more clearly though, and would benefit from an example.

One place to find working examples is in Solr's own test suite. No mention of `facet.pivot.limit` anywhere in the tests or the source code. But they do combine `facet.pivot` with `facet.limit` in the tests.

https://github.com/apache/solr/blob/5bb9ecdb87e21dc912c52945fe027ccc63b47e0a/solr/core/src/test/org/apache/solr/handler/component/DistributedFacetPivotLongTailTest.java#L73-L77

```java
    SolrParams req = params( "q", "*:*", 
                             "distrib", "false",
                             "facet", "true", 
                             "facet.limit", "10",
                             "facet.pivot", "foo_s,bar_s");
```

https://github.com/apache/solr/blob/5bb9ecdb87e21dc912c52945fe027ccc63b47e0a/solr/core/src/test/org/apache/solr/handler/component/DistributedFacetPivotSmallAdvancedTest.java#L114-L116

```java
    facetParams.add(params("facet", "true",
                           "facet.limit", "1",
                           "facet.pivot", "{!stats=s1}place_t,company_t"));
```

https://github.com/apache/solr/blob/5bb9ecdb87e21dc912c52945fe027ccc63b47e0a/solr/core/src/test/org/apache/solr/handler/component/FacetPivotSmallTest.java#L402-L409

```java
      SolrParams p = SolrParams.wrapDefaults(params("q", "*:*",
              "rows", "0",
              "facet", "true",
              "facet.pivot", "company_t",
              "facet.sort", "index",
              "facet.pivot.mincount", "4",
              "facet.limit", "4"),
          variableParams);
```

tsterker:
@thomascorthals Thanks for your details.

Do I understand correctly that you would agree that there is no such thing as `facet.pivot.limit`? This would further supported by me failing to get expected results with `facet.pivot.limit` when directly interacting with Solr.

thomascorthals:
> Do I understand correctly that you would agree that there is no such thing as `facet.pivot.limit`? This would further supported by me failing to get expected results with `facet.pivot.limit` when directly interacting with Solr.

I do agree that there is no such thing as `facet.pivot.limit`.

The `setLimit()`/`getLimit()` methods don't need to be removed though. The parameter should simply be `facet.limit`.

src/Component/Facet/Pivot.php
```diff
@@ -36,28 +36,6 @@ public function getType(): string
...
-    /**
-     * Get the facet limit.
-     *
-     * @return int
-     */
-    public function getLimit(): ?int
-    {
-        return $this->getOption('limit');
-    }
-
```
thomascorthals, line 60: These methods can stay!

src/Component/RequestBuilder/FacetSet.php
```diff
@@ -282,7 +282,6 @@ public function addFacetPivot($request, $facet): void
             )
         );
         $request->addParam('facet.pivot.mincount', $facet->getMinCount(), true);
-        $request->addParam('facet.pivot.limit', $facet->getLimit(), true);
```
thomascorthals, line 285:
```suggestion
        $request->addParam('facet.limit', $facet->getLimit(), true);
```

thomascorthals (changes requested):
You'll have to update the unit test as well.

https://github.com/solariumphp/solarium/blob/fa93d0d5fb5b6797922f4e486e957d789c2bfba3/tests/Component/RequestBuilder/FacetSetTest.php#L374-L377

thomascorthals:
@tsterker Looking into this further, it turned out there were bigger issues with pivot facets than this misnamed parameter. I've opened PR #963 to address them all at once. You're welcome to join the discussion there!

PR #926, opened 2021-04-07
MoreLikeThis overhaul

thomascorthals:
The initial intention was to just add support for the `mlt.maxdf` and `mlt.maxdfpct` parameters. As most parameters are common between the MoreLikeThis Component and QueryType, there was a lot of duplicate code between the two. I decided to move shared code to a trait and only put the few specific parameters in the proper classes.

To facilitate cross-referencing between the code and the [Solr ref guide](https://solr.apache.org/guide/8_8/morelikethis.html), I rearranged the order of the parameters in the source code (setters, request builder …) to match the order in the ref guide.

Support for `mlt.match.include` and `mlt.match.offset` was duplicated in the Component, even though those parameters only work with MLT queries. I've opted to deprecate their setters/getters for now instead of removing them outright because `setMatchInclude()` was used incorrectly in the examples (see below).

~Although the ref guide doesn't list `mlt.interestingTerms` as a common parameter, it is listed for the MLT Component since Solr 8.2 as _Same as defined below for the MLT Handler._~ The ref guide lists `mlt.interestingTerms` (added to the MLT Component in Solr 8.2) as a common parameter since the Solr 8.8.2 release. I've put it in the trait and added support for it in the Component response parser and result + a short explanation in the docs.

The examples for the MLT QueryType were both incorrect, probably due to the fact that techproducts doesn't come with an `/mlt` handler to run them against.

*2.3.1-mlt-query* executed a Select query with an MLT Component instead of an MLT query. It used `setMatchInclude()` on the Component, even though the `mlt.match.include` parameter is only meaningful for MLT queries. I've rewritten it to execute a real MLT query and updated the docs with the working code.

*2.3.2-mlt-stream* tried to display the match document which doesn't exist in that use case. The use case wasn't really clear from the example code either. It looked a lot like a regular query against a default field. I've adapted it to better show that it matches against supplied text instead of a queried document and added a section to the docs about this. (The Solr ref guide uses the same example we had, but it's less confusing with their surrounding prose.)

I've added an `/mlt` handler in `solrconf.patch` and wrote integration tests based on the example scripts for the Component and the QueryType. With this patched config, the examples can also run in the workflow. To allow standalone execution of `examples/execute_all.php`, I also included a check for the handler and an API query to add it if it doesn't exist. The same API query is included in the docs for people who want to run those examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 ### Added
+- MoreLikeThis::setMaximumDocumentFrequency()
+- MoreLikeThis::setMaximumDocumentFrequencyPercentage()
+- getInterestingTerms() of MoreLikeThis Component results
 
 ### Fixed
 
 ### Changed
 
 ### Removed
+
+### Deprecated
+- Support for `mlt.match.include` and `mlt.match.offset` in MoreLikeThis Component (they only work in MLT queries)
 
 
 ## [6.1.1]
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -95,12 +95,13 @@ Overview of Solarium exceptions
 
 ### Logic exceptions
 
-Logic exceptions represent errors in the program logic. You should fix them in your code.
+Logic exceptions represent errors in the program logic. You should prevent them from happening with proper input validation in your code.
 
-Solarium can throw one type of logic exception. It extends its SPL counterpart to implement `Solarium\Exception\ExceptionInterface` and `Solarium\Exception\LogicExceptionInterface`.
+Solarium can throw two types of logic exceptions. Both extend their SPL counterpart to implement `Solarium\Exception\ExceptionInterface` and `Solarium\Exception\LogicExceptionInterface`.
 
 | Exception                                     | Extends                    |
 | --------------------------------------------- | -------------------------- |
+| `Solarium\Exception\DomainException`          | `DomainException`          |
 | `Solarium\Exception\InvalidArgumentException` | `InvalidArgumentException` |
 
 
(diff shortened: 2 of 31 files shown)


thomascorthals:
Quick update on where I'm at: stuck at getting the integration tests to pass on SolrCloud.

For `MoreLikeThisHandler`, I'm inclined to skip them altogether in cloud mode. [SOLR-4414](https://issues.apache.org/jira/browse/SOLR-4414) and [SOLR-5480](https://issues.apache.org/jira/browse/SOLR-5480) make it seem pointless to spend more time on it at this point, especially since [the comments on SOLR-5480](https://issues.apache.org/jira/browse/SOLR-5480?focusedCommentId=17008040&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17008040) mention a possible deprecation of the handler.

For `MoreLikeThisComponent`, the interesting terms are missing from the response in cloud mode. I've inquired on the mailing list whether this is a bug or an undocumented limitation. In either case, I'll have to skip that part of the test, but then I'll know what to put in the comment above that `if`. Or maybe I'm missing something and I will be able to get it to work.

thomascorthals:
> For `MoreLikeThisComponent`, the interesting terms are missing from the response in cloud mode. I've inquired on the mailing list whether this is a bug or an undocumented limitation. In either case, I'll have to skip that part of the test, but then I'll know what to put in the comment above that `if`. Or maybe I'm missing something and I will be able to get it to work.

It's a currently undocumented limitation. Solr's MoreLikeThis documentation has been overhauled recently and this will be released with Solr 8.8.2 very soon. I'll wait for the release and doublecheck our implementation and documentation against the improved ref guide.

thomascorthals:
I've skipped the tests that don't work (or at least not as expected) on SolrCloud with either Solr JIRA issues or the ref guide as justification.

The query examples are now actually run in the workflow thanks to the config patch for the integration tests.

My initial comment is updated to reflect the changes from my last commit.

mkalkbrenner (approved):

PR #930, opened 2021-06-22
fixed return values of Debug Component

mkalkbrenner:
@thomascorthals some function might return null.
I also changed the type of the sub details to be an object but the code should be backward compatible to the old array syntax

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -45,12 +45,12 @@ jobs:
                 ref: branch_7_7
                 path: lucene-solr
 
-            - name: Checkout solr 8.5
+            - name: Checkout solr 8.9
               if: matrix.solr == 8
               uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


thomascorthals:
I'll have to play "Codecov's advocate" here. There's some missing coverage in the tests.

(I assume Codecov doesn't run because it's waiting on the PHP 7.2 tests that are no longer in the workflow. But I have no idea how to change this.)

mkalkbrenner:
I think that I increased the coverage compared to the previous test.

github changed the UI for the settings. I already removed the "required" tests. I assume that this doesn't affect old PRs. But this one is newer :-(

I'll check the settings again.

mkalkbrenner:
I managed to adjust the required checks.

But codecov throws random access denied errors since 10 days.

@thomascorthals I suggest to merge this PR to see if it gets triggered on master.

PR #931, opened 2021-06-23
Full test coverage for Component\Result\Debug

thomascorthals:
@mkalkbrenner 
I've completed the test coverage for Debug. There was a small bug in `Detail::__toString()` when no sub-details were set that I had overlooked without the test.


mkalkbrenner:
@thomascorthals you're right!

I think the error in __toString() isn't critical as it is new.

The codecov integration is broken. It seems to be a permission problem. The webhook gets 403 responses.
I tried to delete the project in codecov.io. But that also leads to a permission denied error message.

There's an option to "reset the webhook". That leads to an error now because the webhook in github is already deleted.

mkalkbrenner (approved):

PR #937, opened 2021-06-26
Bugfix for facet queries with local parameters

thomascorthals:
Closes #229. Closes #936.

I've found a solution for this problem that doesn't break existing code, doesn't sacrifice tagging/excluding as a trade-off and doesn't require any extra effort from the users.

PR #939, opened 2021-06-28
Generic correct merging of local params

thomascorthals:
I took the idea from #937 and applied it to `AbstractRequestBuilder::renderLocalParams()` instead. That should catch all cases where a user puts local params in a query and we try to add more in our code.

I hit a snag with converting code that uses `LocalParameters::render()` though and I would like feedback from @wickedOne. The way parameters are stored and rendered within `LocalParameters` isn't compatible with the key => value array `renderLocalParams()` expects. I have included a kludge to parse them into such an array for now as a proof-of-concept. It works as expected: all tests pass and the output from the examples is identical.

Would it be ok to remove `LocalParameters::render()` completely? Or does it add something that I'm missing?

---

Also added array support to `Helper::qparser()` so it handles `['excludeTags' => ['tag1', 'tag2']]` the same way as `AbstractRequestBuilder::renderLocalParams()` does.


mkalkbrenner (approved):
The patch looks good to me.

mkalkbrenner:
@wickedOne @thomascorthals Should I merge the PR as it is or should we wait for another review?

thomascorthals:
@mkalkbrenner I still had to remove `LocalParameters::render()` but that meant rewriting all the unit tests for `LocalParameters`. I wanted to be sure it was worth the effort first. 😉 Also cleaned up the code to get the actual parameters for rendering, my POC used a kludge for that.

Noticed that booleans weren't yet converted to `'true'` and `'false'` and that `0` wouldn't be rendered while it is a valid value. Fixed those as well.

mkalkbrenner (approved):

PR #943, opened 2021-07-07
Stats overhaul

### LINKED ISSUE #534: How to execute percentiles functionality?
Hello,

Is there code example for Percentiles stats ? As In Stats component, getPercentiles() method there, but how to use this method/related functionality not defined.

In change log ## 3.6.0 - 2016-05-03
     - added: support for additional stats values like percentiles

Can you update example : 2.1.5.10-stats.php ?

### LINKED ISSUE #865: Stats component has "unfriendly" results for percentiles
**Solarium version(s) affected**: 3.6.0+
**Solr version**: 5.1.0+

**Description**
When requesting percentiles for `stats.field`, the returned result is a numeric array with every even index pointing to the requested percentile and the next odd index pointing to the returned value for that percentile.

As tested on `techproducts`:
```
stats.field={!percentiles='1,10,90,99'}popularity

array(8) {
  [0]=>string(3) "1.0"
  [1]=>float(0.14)
  [2]=>string(4) "10.0"
  [3]=>float(1)
  [4]=>string(4) "90.0"
  [5]=>float(8.8)
  [6]=>string(4) "99.0"
  [7]=>float(10)
}

stats.field={!percentiles='99,99.9,99.99'}price

array(6) {
  [0]=>string(4) "99.0"
  [1]=>float(1966.6484985352)
  [2]=>string(4) "99.9"
  [3]=>float(2175.7648498535)
  [4]=>string(5) "99.99"
  [5]=>float(2196.6764849853)
}
```

This requires some "gymnastics" in the user's code to extract the percentiles.

**How to reproduce**  
```php
$query = $client->createSelect();
$query->setRows(0);

$stats = $query->getStats();
$stats->createField('{!percentiles="1,10,90,99"}popularity');
$stats->createField('{!percentiles="99,99.9,99.99"}price');

$resultset = $client->select($query);
$statsResult = $resultset->getStats();

foreach ($statsResult as $field) {
    echo '<h1>' . $field->getName() . '</h1>';
    echo '<table>';
    foreach ($field->getPercentiles() as $i => $percentile) {
        if (0 === $i % 2) {
            echo '<tr><th>' . $percentile . '</th>';
        }
        else {
            echo '<td>' . $percentile . '</td></tr>';
        }
    }
    echo '</table>';
    echo '<hr/>';
}
```

**Possible Solution**  
Return the results as an associative array with the requested percentile as the key pointing to the returned value.

```
stats.field={!percentiles='1,10,90,99'}popularity

array(4) {
  ["1.0"]=>float(0.14)
  ["10.0"]=>float(1)
  ["90.0"]=>float(8.8)
  ["99.0"]=>float(10)
}

stats.field={!percentiles='99,99.9,99.99'}price

array(3) {
  ["99.0"]=>float(1966.6484985352)
  ["99.9"]=>float(2175.7648498535)
  ["99.99"]=>float(2196.6764849853)
}
```

This will break backward compatibility.

**Additional context**  
I was working on an example for #534. I have it working locally with the even/odd kludge but I'm hesitant to open a PR. If we are going to break BC on this, it doesn't help to provide documentation and then break it if it was undocumented to begin with.

thomascorthals:
Closes #534. Closes #865.

Added support for the recent additions to Solr's [Stats Component](https://solr.apache.org/guide/8_9/the-stats-component.html): `distinctValues`, `countDistinct`, `cardinality`.

Moved common functionality between `Result\Stats\Result` and `Result\Stats\FacetValue` to a trait. This avoids inconsistencies like `percentiles` being support in `Result` but not in `FacetValue`. As a consequence, I had to rename the recently-made-public (PR #922) `getValue()` to `getStatValue()` because `FacetValue` already had a public method `getValue()` that can't be renamed without breaking BC. I hadn't anticipated that PR to be merged and released already, so now I have to break BC on that. I think that's the lesser evil because it's so recent?

I don't think there was a way that `Component\Result\Stats\FacetValue::getFacets()` could have ever returned anything so I deprecated it. I guess it was only there because it was copypasted from `Result`, but correct me if I'm wrong!

Inconsistencies from the JSON representation are normalized in the Response Parser.

New functionality is mentioned summarily in the docs and full examples have been added.

### DIFF
--- a/src/Component/Result/Stats/Result.php
+++ b/src/Component/Result/Stats/Result.php
@@ -14,19 +14,14 @@ namespace Solarium\Component\Result\Stats;
  */
 class Result
 {
+    use ResultTrait;
+
     /**
      * Field name.
      *
      * @var string
      */
     protected $field;
-
-    /**
-     * Stats data.
-     *
-     * @var array
-     */
-    protected $stats;
 
     /**
      * Constructor.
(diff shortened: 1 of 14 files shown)


mkalkbrenner (approved):

mkalkbrenner:
> As a consequence, I had to rename the recently-made-public (PR #922) getValue() to getStatValue() because FacetValue already had a public method getValue() that can't be renamed without breaking BC. I hadn't anticipated that PR to be merged and released already, so now I have to break BC on that. I think that's the lesser evil because it's so recent?

I agree!

src/Component/Result/Stats/Result.php
```diff
@@ -14,20 +14,15 @@
  */
 class Result
 {
+    use ResultTrait;
```
mkalkbrenner, line 17: What about re-adding getValue() for BC and declare it as deprecated?

```php
    /**
     * Get value by name.
     *
     * @deprcated
     *
     * @param mixed $name
     *
     * @return string|array|null
     */
    public function getValue($name)
    {
        return $this->getStatValue($name);
    }
```

mkalkbrenner (changes requested):

src/Component/Result/Stats/Result.php
mkalkbrenner: @thomascorthals ^

thomascorthals: That is so obvious that my facepalm must have registered on seismographs. 🤦‍♂️

mkalkbrenner (dismissed):

mkalkbrenner (approved):

PR #948, opened 2021-08-05
Configsets API

mkalkbrenner:


mkalkbrenner:
@thomascorthals I need to add some more test coverage, but it would be nice if you could do a first review of the implementation in general.

The cloud integration tests now use a "real" zookeeper/solr cluster!

mkalkbrenner:
BTW I exceeded our quota at codecov. So the coverage will not be meassured anymore today.

mkalkbrenner:
@thomascorthals I'll merge this PR. If you find something later, we could still improve it in a new PR.
But as I strictly followed your implementation of the collection API I don't see a anything fundamental that will block the PR.

PR #949, opened 2021-08-10
add possibility to set ConnectionTimeout for CurlAdapter

m-h-1:


tests/Integration/SolrServer/CurlTest.php
```diff
@@ -22,6 +22,7 @@ public function setUp(): void
         // Set it to the PHP default of 13s.
         $adapter = new Curl();
         $adapter->setTimeout(CURLOPT_TIMEOUT);
+        $adapter->setConnectionTimeout(CURLOPT_CONNECTTIMEOUT);
```
thomascorthals, line 25: The value of `CURLOPT_CONNECTTIMEOUT` is `78`. That's a bit excessive for a connection timeout. The comment about the "PHP default of 13s" is misleading, that is not what `CURLOPT_TIMEOUT` having a value of `13` means. These values are merely a number by which each option is identified by libcurl internally.

It's probably better to set both options to a reasonable number. 5 seconds should be plenty for the connection timeout. A 15 second request timeout is definitely enough for the current integration tests as we're currently running them within 13 seconds.

thomascorthals (approved):

thomascorthals:
@mkalkbrenner There's no unit test coverage, but `TimeoutAwareTrait` doesn't have it either for the cURL adapter. I can add both afterwards, probably as soon as this week. (I'm currently working on the integration test I promised for `PostBigExtractRequest`, I'll pick it up next.)

PR #954, opened 2021-08-23
Always respect automatic filtering of control characters

### LINKED ISSUE #953: Document fields bypass filterControlCharacters
Hi,

Today we had a hotfix release because of

https://github.com/solariumphp/solarium/blob/beec496540c3d227201c556729d2c61d1c1f8ab1/src/QueryType/Update/Query/Document.php#L158-L160

which bypasses

https://github.com/solariumphp/solarium/blob/beec496540c3d227201c556729d2c61d1c1f8ab1/src/QueryType/Update/Query/Document.php#L149

why? :D

thomascorthals:
Fixes #953.

Made sure control characters are always filtered by default be passing them from the constructor to a new method `setFields()` which ultimately passes them to `setField()` or `addField()` if it's a multivalue field.

Unit tests were expanded to test the filtering when values are set through the constructor, by property or with a method call to ensure the behaviour remains consistent no matter how methods rely upon each other internally.

There is no longer a way to set unfiltered values through the constructor. That could be a BC break for people relying on the old behaviour, but it was never documented to start with.

I had to update an unrelated unit test that passed boosts through the constructor as `int`s. They are now cast to `float` because of the typehint in `setField()` they previously didn't encounter. Solr would do the same casting anyway so this doesn't change what will be indexed.

There was another small inconsistency. When removing a field from a Document, the field's modifier wasn't removed while its boost was. It doesn't matter unless you set a field with the same name again without specifying a modifier. At that point applying the modifier of a removed field to a newly set field seems like a bug to me. (Feel free to disagree with me.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [unreleased]
+### Added
+- QueryType\Update\Query\Document::setFields() to set all fields on a Document
+
+### Fixed
+- Always respect automatic filtering of control characters in field values in QueryType\Update\Query\Document
+
+
 ## [6.1.5]
 ### Added
 - Component\Result\Stats\Result::getDistinctValues()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Always respect automatic filtering of control characters in field values in QueryType\Update\Query\Document
+- Remove the field modifier along with the value(s) and boost in QueryType\Update\Query\Document::removeField()
 
 
 ## [6.1.5]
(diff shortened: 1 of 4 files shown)


thomascorthals:
> There is no longer a way to set unfiltered values through the constructor. That could be a BC break for people relying on the old behaviour, but it was never documented to start with.

If one really must, I suppose something like this will work:

```php
use Solarium\QueryType\Update\Query\Document;
use Solarium\QueryType\Update\Query\Query;

class MyDocument extends Document
{
    protected $filterControlCharacters = false;
}

$query = new Query(['documentclass' => MyDocument::class]);
$doc = $query->createDocument($fields);
```

jupevi:
This patch broke indexing of single nested child documents, turning the associative array into a sequential array by discarding keys here: https://github.com/solariumphp/solarium/pull/954/files#diff-55c0919de6efccffaca331e4cd644e37dd4fd7deae7f17413e21b6320ab36fa3R251-R256

Before this change, this payload worked fine:

```json
{
  "id": 1
  "manufacturer": {
    "docType": "Manufacturer",
    "id": 2,
    "name": "Apple"
  }
}
```

After this change, the payload gets changed to the following before sending to Solr:

```json
{
  "id": 1
  "manufacturer": [
    "Manufacturer",
    2,
    "Apple"
  ]
}
```

Understandably, Solr thinks it's now being fed a multi-valued field instead of a child document, and errors out with `undefined field 'manufacturer'`.

As I was in a hurry to get this working, I did not do a fork to fix this myself, but instead reverted to version 6.1.5 which still works. Just making this comment as a note to myself and for anyone interested.

PR #963, opened 2021-12-10
Fixes and updates for Field, Pivot and JsonTerms facets

thomascorthals:
I was looking into #924 and noticed that there were other request parameters for facets that weren't yet supported. This escalated into rewriting how common methods are reused between facet components. I think my PR follows Solr's intentions more closely and yields a net reduction in duplicate code. However, there's a backward compatibility and some other issues with `facet.pivot.mincount` which I'd like feedback on.

### Doing away with `Component\Facet\AbstractField`

This abstract class was extended by both `Field` and `JsonTerms` so that they could share methods that seemingly set "the same" parameter. Those parameters have the same name, but there is a distinction. The parameters for JSON facets deal with buckets instead of fields. This may seem like a pedantic matter of semantics, but the values for `sort` are incompatible between the two. Trying to reuse `setSort()` meant the PHPDoc actually instructed users to use wrong values for `JsonTerms`. It seems more natural to me to keep the methods for JSON facet parameters separate from the non-JSON ones to avoid this type of confusion.

### Updating `Component\Facet\JsonTerms`

Moved constants, properties and methods from `AbstractField` to `JsonTerms`. Wrote correct PHPDoc for `setSort()`. Deprecated `SORT_COUNT` and `SORT_INDEX` because they're wrong. Adding `SORT_COUNT_ASC`, `SORT_COUNT_DESC`, `SORT_INDEX_ASC`, `SORT_INDEX_DESC` was actually an afterthought when I was writing the CHANGELOG. They provide something to point to as a replacement for the deprecated constants and they make it obvious to users what was wrong the original ones. Passing a literal value like `'index asc'` did already work and doesn't break.

While I was at it, I also rounded out support for all current parameters.

### Introducing `Component\Facet\FieldValueParameters{Interface,Trait}`

The ref guide for [Field-Value Faceting Parameters](https://solr.apache.org/guide/8_11/faceting.html#field-value-faceting-parameters) says:
> Unless otherwise specified, all of the parameters below can be specified on a per-field basis with the syntax of `f.<fieldname>.facet.<parameter>`

This means that the parameters with identical names in `Component\FacetSet` and `Component\Facet\Field` are in fact the same parameters "in Solr's mind". It makes sense for them to share the same methods and constants. I opted for a trait because I didn't want to mess with the inheritance and had to bring an interface into the mix because traits can't define `const`s.

All current parameters are now supported.

`METHOD_FCS` is added for per-segment field faceting for single-valued string fields.

`METHOD_UIF` is added for UnInvertedField faceting. Although not mentioned in the ref guide, it was added in [SOLR-8466](https://issues.apache.org/jira/browse/SOLR-8466).

https://github.com/apache/solr/blob/470b7fbff96aec0bd1120634f3f8a8241484f6f7/solr/solrj/src/java/org/apache/solr/common/params/FacetParams.java#L58-L61

```java
  /**
   * Value for FACET_METHOD param to indicate that Solr should use an UnInvertedField
   */
  public static final String FACET_METHOD_uif = "uif";
```

### Fixing `Component\Facet\Pivot`

`Pivot` supported two parameters, but both of them had issues with the implementation.

The ref guide for [Additional Pivot Parameters](https://solr.apache.org/guide/8_11/faceting.html#additional-pivot-parameters) says:

> Although `facet.pivot.mincount` deviates in name from the `facet.mincount` parameter used by field faceting, many of the faceting parameters described above can also be used with pivot faceting:
>
> * `facet.limit`
> * `facet.offset`
> * `facet.sort`
> * `facet.overrequest.count`
> * `facet.overrequest.ratio`

This doesn't mention explicitly that these parameters can also be specified on a per-field basis. Digging through the Solr code, I found that they can.

https://github.com/apache/solr/blob/470b7fbff96aec0bd1120634f3f8a8241484f6f7/solr/core/src/test/org/apache/solr/handler/component/DistributedFacetPivotWhiteBoxTest.java#L77-L91

```java
    SolrParams params = params("facet", "true", 
                               "q", "*:*", 
                               // "wt", "javabin", 
                               "facet.pivot", "{!stats=s1}place_t,company_t", 
                               // "version", "2", 
                               "start", "0", "rows", "0", 
                               "fsv", "true", 
                               "fl", "id,score",
                               "stats", "true", 
                               "stats.field", "{!key=avg_price tag=s1}price_ti",
                               "f.place_t.facet.limit", "160", 
                               "f.place_t.facet.pivot.mincount", "0", 
                               "f.company_t.facet.limit", "160", 
                               "f.company_t.facet.pivot.mincount", "0", 
                               "isShard", "true", "distrib", "false");
```

This raises a problem with our current implementation. If `mincount` is set on a pivot facet, the request builder adds a general `facet.pivot.mincount` parameter. If it's set on multiple pivot facets, the previous ones are overwritten and the value of the last one will be the value for all pivots. It can't be set per-field. I have tried to figure it out, but I can't seem to get it 'perfect'.

I don't know if that test is meant to imply that the per-field params always have to be the same for all the fields listed in a pivot. My proposed solution assumes it is. Assuming otherwise would mean added complexity and I don't know if there's a use case to justify that.

#### `setMinCount()`

This proposal deprecates `Pivot::setMinCount()` in favour of `FacetSet::setPivotMinCount()` or `Pivot::setPivotMinCount()`. This makes it very clear that something was up with the old method and a conscious choice has to be made by the user on global versus specific params. It also uses the config option `pivot.mincount` because `mincount` already has a different meaning for `FacetSet`s. When it came to dealing with config mode, I started second-guessing this approach. Truly deprecating the old way would mean having to deprecated `mincount` used as a config option. This would have to be done with `trigger_error()` in `Pivot::init()` and I'm hesitant because there's no precedent for this approach in the codebase. I dislike the way this scenario breaks BC because the parameter stops being included in the request at all, changing the outcome of existing code.

Another solution would be to keep `Pivot::setMinCount()` and the `mincount` config option as a way to set the per-field param and use `FacetSet::setPivotMinCount()` and `pivot.mincount` for the general param. I dislike two things about that scenario. It changes the meaning of `Pivot::setMinCount()` behind the scenes without alerting the user. It feels inconsistent to have differently named methods on `FacetSet` and `Pivot` to set essentially the same parameter.

A possible middle ground would be to use `setPivotMinCount()`/`pivot.mincount` as the 'proper' way to set the parameter and keep the current `Pivot::setMinCount()` implementation as it is, but mark it as `@deprecated`. I dislike that this keeps the buggy behaviour intact and doesn't solve anything for users who understandably assumed it did work per-field already. I also dislike that it would introduce some complexity in the request builder when it is combined with `FacetSet::setPivotMinCount()`.

The main question boils down to this: do we prefer a correct and unambiguous implementation at the cost of a hard BC break? Or do we prefer a frictionless update path at the cost of keeping a demonstrably buggy implementation around?

#### `setLimit()`

The same mix-up between general and per-field parameters was made with `Pivot::setLimit()`. As an addtional bug, this method resulted in setting a `facet.pivot.limit` parameter on the request. Per the ref quide quoted above, this parameter should in fact be `facet.limit`. Since it has never worked anyway, I went ahead and implemented it the 'right' way for general and per-field with the correct parameter. It will change the outcome for existing code in that it will start working properly.

#### Additional parameters

The other parameters are now also supported in the same way as `limit`.

### Moving code around

I've followed the same order in which parameters are listed in the ref guide throughout the codebase: setters, request builder, unit tests, docs.


thomascorthals:
@mkalkbrenner @wickedOne If you find the time, you should read the full story of how I went from "there is no `facet.pivot.limit` parameter in Solr" to this 1.8 kloc PR. It's been a rabbit hole so do point out any conceptual or implementation mistakes I've made along the way.

The TL;DR version of why I've marked it as a draft is two-fold.

1. Is it a safe assumption that per-field parameters on pivot facets can be set to the same value for every field in the facet? I don't have real-life use cases to prove or disprove the validity of that assumption.

2. How should we deal with transitioning from `Pivot::setMinCount()` setting the general parameter to setting it per-field? I went back and forth between doing it properly with a hard BC break (the current state of the PR) and holding on to the old way which would add complexity to the codebase. Also, how should we deprecate the config option in `Pivot::init()` if the corresponding setter gets `@deprecated`?

mkalkbrenner:
@thomascorthals first of all, great work!

Unfortunately I don't use pivots in any of my projects.
But if we take into account that the current implementation is considered to be broken or at least incomplete and erroneous, I think we can be radical and introduce a BC break instead of more complex code.
We could release this as  6.2.0 beta.1 or rc.1 and wait for some feedback. But to be honest I don't expect to get feedback on this until someone runs into an issue with the final release.

BTW the brnach has a confict with master.

mkalkbrenner (approved):

PR #967, opened 2021-12-24
Possible to set non-default context on an endpoint

### LINKED ISSUE #853: Configure endpoint for proxied Solr
We keep getting issues about accessing Solr on other paths than `/solr` (#787, #815, #852, #938).

I was going to provide an example and doc note of the solution I came up with in #852 for accessing legacy Solrs "at your own risk": extend the cURL adapter and change the path on the returned handle. When trying to pinpoint the Solr version that had the path hardcoded (or maybe it's always been like that), I stumbled upon this optimisation in Solr 6.5.0:
```markdown
* SOLR-9584: Support Solr being proxied with another endpoint than default /solr, by using relative links
  in AdminUI javascripts (Yun Jie Zhou via janhoy)
```

This seems like a legitimate reason for having the endpoint configurable. What is the best way to support this?

1. Include my "hack" in the docs as the solution for proxied Solrs. Not very user-friendly if it's more than a mere courtesy for legacy installations.
2. Make it configurable on `Endpoint`. Must still default to `/solr` as most users won't have to bother with it.
3. Make it possible via the `CustomizeRequest` plugin. It is a customisation on the request of sorts after all.
4. Write a dedicated `ProxiedEndpoint` plugin.

I haven't found any mentions of proxying `/api`. I guess that would also be possible?

thomascorthals:
Closes #853.

I've assumed in README/getting-started that this will be released in Solarium 6.3. Looks 'cleaner' than 6.2.1 if it has to live permanently in the documentation.

---

FYI, writing unit tests like this doesn't work as was apparently expected.

https://github.com/solariumphp/solarium/blob/f63861d9246e8f06487ec2dfaea3cef0d4babccc/tests/Core/Client/EndpointTest.php#L93-L101

The unit test ends when the exception is encountered and the second assertion is never tested. One solution would be to have the test for the exception as the last one in the method. For clarity, I put the exception tests in their own methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [6.3.0]
+### Added
+- Possibility to set the context on an endpoint for SolrCloud instances with a non-default `hostContext` or Solr instances behind a reverse proxy, defaults to `solr` if omitted
+
+
 ## [6.2.0]
 ### Added
 - Component\FacetSet::setOffset()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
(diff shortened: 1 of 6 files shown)


mkalkbrenner:
Once again, thanks a lot for your work!

I think that this doesn't cause BC breaks because the correct default value is provided.
So why should we postpone this feature. I recommend to commit it right away and release it with the next patch release.

thomascorthals:
I've changed it to 6.2.1.

Issue #523, opened 2017-09-15
BufferedAdd with many documents slowly increases flush cycle time


rms2219:
I'm using the BufferedAdd plugin to index my documents. In some instances, I'll need to index tens of millions of documents. While testing, I set my buffer size to 30,000. I noticed that when my indexing process begins, the flushing those 30K documents takes about a second. After a couple million documents, that cycle takes a little longer (~1 second more). This goes on and on, and at around 10 or 11 million documents that have been indexed, I'm up to like 11 seconds between cycles. Is there something that Solarium is holding onto or writing to that would be slowing this down? I should add that this happens regardless of whether the Solr index is empty or whether it's already populated with millions of other documents.

thomascorthals:
I have tried to reproduce this with a buffer size of 30,000 and a dataset of 12,000,000 documents against a local Solr on my laptop. There is a noticeable difference between smaller (< 100 KiB) and larger (~ 600 KiB) docs. There are periodical peaks in the time between flushes that I didn't look into because they're not what this issue is about.

![graph](https://user-images.githubusercontent.com/12425704/90333151-9fc00600-dfc3-11ea-98fd-8808f5bc72de.png)

The trendlines are near horizontal. There is a small upward slope (<i>m</i><sub>Small docs</sub> = 0.0002 and <i>m</i><sub>Large docs</sub> = 0.0003), but not nearly as much as @rms2219 was experiencing. My document ids were generated as `test_1 .. test_12000000` so the later buffers are almost 90 KiB larger than the first ones. I tried again with `test_00000001 .. test_12000000` to keep the buffer size consistent throughout the test. This time, the slope was slightly downward (<i>m</i><sub>Small docs</sub> = −0.000009 and <i>m</i><sub>Large docs</sub> = −0.0002).

![graph](https://user-images.githubusercontent.com/12425704/90342045-555f7900-e005-11ea-859a-d966e2f2f65c.png)

I think it's safe to assume that larger buffers lead to slower flushes if everything else remains the same. One thing that could cause a steadily growing delay is a loop bleeding over data in the next iteration. I tried this:

```php
set_time_limit(0);
ini_set('memory_limit', ini_get('suhosin.memory_limit') ?: -1);

$data = [
    'name' => 'Test doc',
    'ignored_cat' => [],
];

for ($i = 0; $i < 12000000; ++$i) {
    $data['id'] = 'test_'.$i;
    $data['ignored_cat'][] = $i;
    $buffer->createDocument($data);
[...]

PR #971, opened 2022-01-03
Add BufferedDelete plugin

### LINKED ISSUE #495: Feature Request: Add Deletes to BufferedAdd plugin
Ability to add a delete (by id or by query) to the 'BufferedAdd' buffer would be incredibly helpful when trying to reduce the number of commits sent to Solr.

For example, if I have a cron that performs a commit every 5 minutes, all of my adds/updates that have been sent to Solr are included in that. However when I do a delete, that appears to require a commit each time I do a check for new or updated items to buffer, which increases the number of commits performed in general.

thomascorthals:
Closes #495.

~~**BIG DISCLAIMER:** I haven't benchmarked this yet. Let's see if the code stands up to scrutiny first.~~

There is some additional housekeeping involved to be able to mix documents and deletes in one buffer and have them sent to Solr in the order they were added(*). This is the most streamlined way I could come up with. If performance for adding only documents suffers from these changes, I could try and work around that with some additional housekeeping and code complexity.

@wickedOne You've mentioned a ~28M document index in #523. Any chance you can run this through my code for a speed comparison? With any luck that issue might also disappear, but I was never able to reproduce it so I can't say.

(*) Actually, if you mix ids and queries, they have always been grouped by type in `Solarium\QueryType\Update\RequestBuilder::buildDeleteXml()`. Within a single delete operation, this doesn't matter for the end result.


mkalkbrenner:
@wickedOne do you plan to test this PR?

wickedOne:
yes i do, will have a go at it this friday

wickedOne (commented):
hi thomas

first of all, i no longer work at the company whose dataset i used and unfortunately the database i still had on my private latop was quite outdated so wasn't able to run my tests on 28m documents. instead i ran them on a core having 1128364 (real life) documents.

### details
- **number of fields**: 49 regular fields, 13 multi value and 9 copy fields
- **buffer size**: 10000
- **number of documents**: 1128364 
- **solr version**: 8.3.1
- **solarium version**: 6.1.1 used for comparison with this branch

### tests
i ran several tests using both 6.1.1 from solarium and this branch, averaged running time and calculated the difference.

the first batch of tests was run using a [`NoopEventDispatcher`](https://gist.github.com/wickedOne/f7b824f73699e7553431586a8d243269). which resulted in a performance hit of 0.05%

the second batch of tests i ran with the symfony event dispatcher which resulted in a performance hit of 3.46%
this is of course without actually using the newly added events

so given the example in #523, where the running time was 45 min for a rebuild, would result in a running time of approx 47 min which probably isn't a fair comparison as i've always explicitly disabled the event dispatcher in batch operations as it gives way too much overhead.

@thomascorthals  @mkalkbrenner not sure whether either of you use events in your batch operations. i would be fine with a performance hit of 0.05% when not using the event system, not sure whether i would appreciate a 3.46% performance hit especially for large cores.

### note
i have one remaining option to get my hands on a more up to date database, but that is probably going to cost me quite a bit of time getting that up and running properly so please let me know whether or not this answers your questions

thomascorthals:
I don't use events at all. But I assume the newly added events shouldn't matter as they're not fired when only documents are added.

This version of the code is already a few iterations in. I was juggling arrays around as "subbuffers per type" until I trimmed that down to using the actual `Command`s as subbuffers. That's as much middle man as I could cut out in that regard.

One thing I could do is adding two flags `$hasAdd` and `$hasDelete` that are set to `true` whenever (and every time) a doc or delete is added and set to `false` when the buffer is cleared. If only of them is true, it could shortcut to an optimised version of `addBufferToQuery()`.

Another possibility is moving deletes to a dedicated `BufferedDelete` plugin. If a user does want to intersperse, it becomes their responsibility to flush the buffer if the order is important.

In either case, I would still use the `AddCommand` directly for documents to see how it plays out in the benchmark and if it could be a positive change for #523.

wickedOne:
yes, i like this approach

will start running the benchmarks and will get back with the results probably somewhere this evening (they take a while)

wickedOne (approved):
ok done running the benchmarks with the same dataset and specs as mentioned in my [previous comment](https://github.com/solariumphp/solarium/pull/971#pullrequestreview-853524947).

with a `NoopEventDispatcher` there's a performance hit of 0.03%
with a functioning EventDispatcherInterface the performance hit is 1.96%
so that's quite an improvement compared to your previous implementation!

the only thing i'm wondering about is whether we should mention a bit in the documentation about the overhead of using the event dispatcher in batch operations, or maybe even add the `NoopEventDispatcher` to the codebase so others can use that implementation to more or less disable it: what do you think?

thomascorthals:
I'm surprised that there's still a performance hit. The updated implementation of `flush()` should be more performant as it now bypasses everything that happens in `QueryType\Update\Command\Add::addDocuments()`.

I'll look into profiling the original `flush()` against my latest version (without the newly introduced inheritance from an abstract superclass) and against this branch to pinpoint where it takes that hit.

wickedOne:
i'll do some additional investigation as well 
i can run a limited set of additions in both branches with a blackfire profile and see what's that difference.
keep in mind i run these benchmarks on my laptop which is not an isolated environment, but assumed running each test run four times and take the averages would give a realistic perspective

thomascorthals:
I never use events, so this would work for me:

```php
    public function __construct(AdapterInterface $adapter, ?EventDispatcherInterface $eventDispatcher = null, array $options = null)
    {
        $this->adapter = $adapter;
        $this->eventDispatcher = $eventDispatcher ?? new NoopEventListener();

        parent::__construct($options);
    }
```

There's another way to eliminate the overhead of events completely: an 'eventless' and an 'eventful' version of the plugins. One version could extend the other and only has to override the add, flush and commit methods. Ideologically, I'd say eventful should inherit from eventless. BC-ologically, it might have to be the other way around because it is eventful now.

thomascorthals:
I've added a simple benchmark to the examples (excluded from the workflow) that runs 1.2M documents through the plugins with different buffer sizes. I wanted to include an unbuffered baseline, but with no buffer or a very small buffer I kept running into `HTTP request failed, Failed to connect to 127.0.0.1 port 8983: Address already in use`. Had to make the delete buffer even larger to avoid this.

Ran this on my laptop as well. For each variant, I started a fresh Solr 8.11.1 with a pristine techproducts core. The same variant was repeated 4 times in succession against that core. Swapped the buffered deletes for a single delete query on the other branches.

### Current `master` branch unaltered

buffer size | run 1 | run 2 | run 3 | run 4 | avg
---|---|---|---|---|---
60 | 393336 ms | 446461 ms | 341250 ms | 342715 ms | 381 s
600 | 128235 ms | 133470 ms | 115498 ms | 117010 ms | 124 s
6000 | 107750 ms | 105054 ms | 102506 ms | 103820 ms | 105 s
60000 | 104365 ms | 103022 ms | 100282 ms | 101162 ms | 102 s
600000 | 107445 ms | 108880 ms | 103414 ms | 110605 ms | 108 s

### Current `master` with the new `flush()` implementation

buffer size | run 1 | run 2 | run 3 | run 4 | avg
---|---|---|---|---|---
60 | 328771 ms | 320223 ms | 327632 ms | 341922 ms | 330 s
600 | 112969 ms | 120271 ms | 117505 ms | 115224 ms | 116 s
6000 | 102331 ms | 104862 ms | 104241 ms | 103915 ms | 104 s
60000 | 99939 ms | 100891 ms | 101621 ms | 100518 ms | 101 s
600000 | 108973 ms | 103619 ms | 103907 ms | 103447 ms | 105 s

### My `buffereddelete` branch

buffer size | run 1 | run 2 | run 3 | run 4 | avg
---|---|---|---|---|---
60 | 330941 ms | 331787 ms | 327259 ms | 336370 ms | 332 s
600 | 113571 ms | 114938 ms | 112458 ms | 121695 ms | 116 s
6000 | 105661 ms | 103240 ms | 102755 ms | 108752 ms | 105 s
60000 | 99869 ms | 100790 ms | 99880 ms | 100363 ms | 100 s
600000 | 102520 ms | 103466 ms | 102286 ms | 105068 ms | 103 s

According to these measurements, the new `flush()` implementation narrowly beats the current one and the effects of inheriting from an abstract base class are inconclusive. Of course, this is not real life data.

Regardless of which implementation makes the final cut, this will be an interesting benchmark to repeat on an eventless `BufferedAddLite` plugin. 😉

thomascorthals:
As expected, there's a noticeable performance gain when the plugins don't dispatch any events.

Results from a single run of `examples/7.5.3-plugin-bufferedupdate-benchmarks.php` on my laptop.

### BufferedAdd / BufferedDelete

add buffer size | add time | delete buffer size | delete time
--- | --- | --- | ---
600 | 123223 ms | 2400 | 26220 ms
6000 | 104771 ms | 24000 | 22761 ms
60000 | 106475 ms | 240000 | 22532 ms
600000 | 112684 ms | 1200000 | 22215 ms

### BufferedAddLite / BufferedDeleteLite

add buffer size | add time | delete buffer size | delete time
--- | --- | --- | ---
600 | 119962 ms | 2400 | 20007 ms
6000 | 101502 ms | 24000 | 15719 ms
60000 | 99202 ms | 240000 | 15021 ms
600000 | 102488 ms | 1200000 | 17227 ms

PR #987, opened 2022-03-09
Fix PHP 8.1 deprecations

### LINKED ISSUE #986: PHP 8.1: Return type deprecation notices in Facet.php & Result.php
When executing every file with:
```
docker run -it --rm php:8.1 /bin/bash
apt update && apt install -y git
git clone https://github.com/solariumphp/solarium.git
cd solarium/src/
find -type f -name \*.php -exec php -e {} \; | grep Deprecated | sort -u
```

following deprecation notices occur:

- Component/Result/Analytics/Facet.php on line 110
Deprecated: Return type of Solarium\Component\Result\Analytics\Facet::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

- Component/Result/Analytics/Result.php on line 78
Deprecated: Return type of Solarium\Component\Result\Analytics\Result::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

thomascorthals:
Fixes #986.

### DIFF
--- a/src/Component/Result/Analytics/Facet.php
+++ b/src/Component/Result/Analytics/Facet.php
@@ -107,7 +107,7 @@ class Facet implements \IteratorAggregate, \Countable
     /**
      * {@inheritdoc}
      */
-    public function getIterator()
+    public function getIterator(): \ArrayIterator
     {
         return new \ArrayIterator($this->results);
     }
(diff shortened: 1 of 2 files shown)


alexgit2k (approved):
Looks good, no deprecation notices anymore.

PR #990, opened 2022-04-07
Fill in some missing test coverage

thomascorthals:
Started with `vendor/bin/phpunit --exclude-group integration` and wrote as many unit tests as was feasible. Seems trivial for setters/getters, but I discovered an actual bug this way.

- 1 bugfix
- some corrected PHPdoc
- minor refactoring
- renamed a few misnamed test methods
- moved all the tests for `src/Component/` to `tests/Component/`


src/Core/Client/Client.php
```diff
@@ -1292,75 +1292,75 @@ public function createCoreAdmin(array $options = null): CoreAdminQuery
     }
 
     /**
-     * Create a managed resources query instance.
+     * Create a Collections API query instance.
```
thomascorthals, line 1295: These methods were moved around to keep the `QueryType\Server` methods grouped together.

src/QueryType/Server/Collections/Query/Action/Create.php
```diff
@@ -294,20 +293,6 @@ public function setAutoAddReplicas(bool $autoAddReplicas): self
...
-     * @param string $id
-     *
-     * @return self Provides fluent interface
-     */
-    public function setAsync(string $id): AsyncActionInterface
-    {
-        $this->setOption('async', $id);
-
-        return $this;
-    }
```
thomascorthals, line 309: This overrides the inherited method with the parameter name as the only difference. None of the other classes that inherit it do this.

https://github.com/solariumphp/solarium/blob/3becc5d3f98501de99d4b993f768687d0cee6b28/src/QueryType/Server/Query/Action/AbstractAsyncAction.php#L20-L25

PR #994, opened 2022-04-28
Correctly handle nested child documents

### LINKED ISSUE #992: Update query broken when document contains single nested child documents
**Solarium version(s) affected**: 6.1.6
**Solr version**: 8.5.2

**Description** 

_Originally posted by @jupevi in https://github.com/solariumphp/solarium/issues/954#issuecomment-1095214385_

Pull request https://github.com/solariumphp/solarium/pull/954 broke update queries containing documents with single nested child documents (i.e. associative arrays), because it always treats arrays in field values as sequential and discards their keys.

Relevant lines in the pull request: https://github.com/solariumphp/solarium/pull/954/files#diff-55c0919de6efccffaca331e4cd644e37dd4fd7deae7f17413e21b6320ab36fa3R251-R256

Before this change, this payload worked fine:

```json
{
  "id": 1,
  "manufacturer": {
    "docType": "Manufacturer",
    "id": 2,
    "name": "Apple"
  }
}
```

After this change, the payload gets changed to the following before sending to Solr:

```json
{
  "id": 1,
  "manufacturer": [
    "Manufacturer",
    2,
    "Apple"
  ]
}
```

**How to reproduce** 

```php
require __DIR__ . '/vendor/autoload.php';
$data = json_decode('{
    "id": 1,
    "manufacturer": {
        "docType": "Manufacturer",
        "id": 2,
        "name": "Apple"
    }
}', true);
$doc = new \Solarium\QueryType\Update\Query\Document($data);
var_dump($doc->getFields());
```

Output:

```
array(2) {
  ["id"]=>
  int(1)
  ["manufacturer"]=>
  array(3) {
    [0]=>
    string(12) "Manufacturer"
    [1]=>
    int(2)
    [2]=>
    string(5) "Apple"
  }
}
```

As you can see, the array keys have been discarded.

Before version 6.1.6 the output was correct:

```
array(2) {
  ["id"]=>
  int(1)
  ["manufacturer"]=>
  array(3) {
    ["docType"]=>
    string(12) "Manufacturer"
    ["id"]=>
    int(2)
    ["name"]=>
    string(5) "Apple"
  }
}
```

thomascorthals:
Fixes the `Document` related part of #992. Documents the known limitation that it won't work with our XML request builder.

(Note: edited to reflect additional commits after I opened the PR.)

`setField()` no longer loses the array keys. `addField()` always puts a nested child document in an array upon the first call. This avoids unpleasant surprises if it could be called one or more times in a loop when you always intend it as multivalued. Solr does this automatically for regular fields if you index a single value for a multivalued field, but nested children are pseudo-fields that aren't defined in the schema.

```php
foreach ($topic->getReactions() as $reaction) {
    $doc->addField('reactions', $reaction);
}
```

Unit and integration tests have been expanded to cover the possible ways to add/set child documents. Documentation including an example has been added.

Moved the control character filtering to the request builder because it's XML specific. It didn't really belong in `Document` anyway and it made things more complicated there. I no longer see a use case for disabling it now that it is in the right place. It is still possible to override if you really want.

```php
use Solarium\Core\Query\Helper;
use Solarium\Core\Query\RequestBuilderInterface;
use Solarium\QueryType\Update\Query\Query as UpdateQuery;
use Solarium\QueryType\Update\RequestBuilder as UpdateRequestBuilder;

$update = new NonControlCharFilteringUpdateQuery();
$doc = $update->createDocument($data);
$update->addDocument($doc);

// and here you'll get HttpException if $data contains control characters
$client->update($update);

class NonControlCharFilteringHelper extends Helper
{
    public function filterControlCharacters(string $data): string
    {
        return $data;
    }
}

class NonControlCharFilteringUpdateQuery extends UpdateQuery
{
    /**
     * Get a requestbuilder that doesn't filter control characters.
     *
     * @return UpdateRequestBuilder
     */
    public function getRequestBuilder(): RequestBuilderInterface
    {
        return new NonControlCharFilteringUpdateRequestBuilder();
    }
}

class NonControlCharFilteringUpdateRequestBuilder extends UpdateRequestBuilder
{
    /**
     * Get a custom helper instance that doesn't filter control characters.
     *
     * @return Helper
     */
    public function getHelper(): Helper
    {
        if (null === $this->helper) {
            $this->helper = new NonControlCharFilteringHelper();
        }

        return $this->helper;
    }
}
```


thomascorthals:
> There is no such thing as a "single nested document" in Solr. It always expects an "array of child documents", JSONly speaking.

At least that's what the ref guide led me to believe. But it actually works when you do it in JSON.

But I can't figure out how to it in XML.

thomascorthals:
> I ask myself if we should switch to JSON entirely when sending docs to Solr. XML is from the early days and we never changed it as it is still supported.

I have been wondering the same thing. But for now, I've focused on fixing `Document`.

I've raised [SOLR-16183](https://issues.apache.org/jira/browse/SOLR-16183) for the XML limitation and listed it as a known limitation for now in the docs.

thomascorthals:
Codecov keeps complaining because I've removed more covered code than I've added so the overall coverage is down by the net difference in total lines. That shouldn't really count as a failure if the diff coverage is 100% imho.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Solarium\QueryType\Server\Collections\Query\Action\ClusterStatus::getRoute() always returned NULL even if a route was set
+- Adding nested child documents through `Document::setField()` and `Document::addField()`
```
mkalkbrenner, line 25: Needs to be moved

CHANGELOG.md
```diff
@@ -30,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 - Solarium\QueryType\Stream\Expression, use Solarium\QueryType\Stream\ExpressionBuilder instead
+- Solarium\QueryType\Update\Query\Document::setFilterControlCharacters(), extend Update\Query\Query to use a custom request builder & helper if you don't want control characters filtered
```
mkalkbrenner, line 34: Needs to be moved

mkalkbrenner (changes requested):

PR #996, opened 2022-05-13
test against Solr 9

mkalkbrenner:


thomascorthals:
@mkalkbrenner I was already preparing that too: https://github.com/solariumphp/solarium/compare/master...thomascorthals:integration?expand=1

But I'm waiting for the image to drop on Docker Hub.

thomascorthals:
I suspect you'll need this change because the Unified Highlighter won't highlight **power** without **cord**.

```diff
diff --git a/tests/Integration/AbstractTechproductsTest.php b/tests/Integration/AbstractTechproductsTest.php
index 769763dc3ec..85c36d06391 100644
--- a/tests/Integration/AbstractTechproductsTest.php
+++ b/tests/Integration/AbstractTechproductsTest.php
@@ -522,7 +522,7 @@ abstract class AbstractTechproductsTest extends TestCase

         $select->setQuery('power cord');
         // Activate highlighting.
-        $select->getHighlighting();
+        $select->getHighlighting()->setMethod(Highlighting::METHOD_ORIGINAL);
         $facetSet = $select->getFacetSet();
         $facetSet->createFacetField('stock')->setField('inStock');
```

thomascorthals:
https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html#deprecations-and-removals
> VelocityResponseWriter is an independent project now; it is no longer a part of Solr. This encompasses all previously included `/browse` and `wt=velocity` examples.

The `/browse` handler is used for a couple of integration tests. We could either:

- have separate solrconf patches for 7/8/9 to reinstate it in 9;
- define our own handler with the necessary components and use that instead for all versions.

thomascorthals:
There's another error with an atomic update of the `price` field. It looks like it has something to do with the `copyField` stuff that ensues. Irrelevant to what we're actually testing on our side. We could change the test to use the `weight` field instead.

thomascorthals:
@mkalkbrenner I'v got the integration tests sorted out in my branch: https://github.com/thomascorthals/solarium/tree/integration

It's stuck on running the examples so I can't open a PR yet.

PR #998, opened 2022-05-14
Test against Solr 9

thomascorthals:
Closes #996. ;-)

Things mostly worked out-of-the-box.

Solr 9 no longer has a `/browse` handler, I replaced it — for all versions — with a self-defined `/componentdemo` handler that has the components we need. That way the test results aren't dependent on the default parameters that were previously present.

There's [one error](https://github.com/thomascorthals/solarium/runs/6433745092?check_suite_focus=true) I couldn't figure out when setting up the examples. I've skipped the examples that use `/mlt` against Solr 9 for now. Will see if I can figure it out later.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -15,7 +15,7 @@ jobs:
         strategy:
             matrix:
                 php: [7.3, 7.4, 8.0, 8.1]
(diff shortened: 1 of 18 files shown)


mkalkbrenner:
So I can close the other branch where I did my experiments with Solr 9

PR #1007, opened 2022-06-17
Cleanup for cursorMark

thomascorthals:
Setting `'cursormark'` in config mode doesn't require special treatment in `init()`.

Also renamed all of the `*Cursormark()` methods to `*CursorMark()` for consistency with Solr's `cursorMark` parameter and the `Result::getNextCursorMark()` method. Doesn't break BC because PHP is case-insensitive about method names.


thomascorthals:
Same "failure" I've experienced elsewhere: code that was covered is removed and Codecov is unhappy about it.

PR #1016, opened 2022-07-15
ParallelExecution dispatches all Execute events

### LINKED ISSUE #1015: ParallelExecution doesn't honour PostBigRequest
### Discussed in https://github.com/solariumphp/solarium/discussions/1014

<div type='discussions-op-text'>

<sup>Originally posted by **leuchtdiode** July 14, 2022</sup>
Hi,

I'm using the PostBigRequest plugin in my Solarium client which is working fine with my "miles long queries".

But it looks like in combination with ParallelExecution the plugin gets omitted?

The result after executing in parallel throws "Request URI too long".

Is it possible to combine these two plugins?

Thx & BR,
Alex</div>

thomascorthals:
Closes #1014. Fixes #1015.

`ParallelExecution` didn't dispatch `PreExecute`, `PreExecuteRequest`, `PostExecuteRequest`, `PostExecute` events. As a result, plugins like `PostBigRequest` couldn't be used in combination with it. I've added those events in the same order they're dispatched for normal "serial" execution.

`ParallelExecution` also threw away timeout/connectiontimeout options because it always created a new Curl adapter on initialisation. Changed that to keep the adapter if it already is Curl, and set the timeout options of the old adapter on the newly created Curl adapter otherwise.

Integration tests were added to cover all possibilities for `ParallelExecution` (including the combination with `PostBigRequest`) and to test that all events are dispatched in the expected order for a normal and single "parallel" request.

As a side note, changed the test query in the `PostBigRequest` example to generate a huge querystring without tiptoeing so close to the maxBooleanClauses limit.


thomascorthals:
I didn't even realise at first … By not replacing an existing cURL adapter, proxy settings are now kept intact on parallel requests. They can't be carried over other adapters though. I've adjusted the docs to make this clear.

thomascorthals:
Docs claim all query types are supported, but Server queries didn't work. This was just a typehinting mixup though. Fixed it and added it to the integration test.

leuchtdiode:
Thank you :) Is there a "time table" for the library when the fix will be released in a new version?

Thx & BR,
Alex

PR #1017, opened 2022-07-19
Check release version

thomascorthals:
Check if `Client::VERSION` matches the tag of a release. Will inform of a mismatch post factum.


thomascorthals:
Not much use for this PR. Its intention is already covered in #1019 with `ClientTest::test76_5_4`.

```php
    public function test76_5_4()
    {
        $this->assertTrue(
            // 76.5.4 is the version tag we use within github actions.
            Client::checkExact('76.5.4')
        );
    }
```

PR #1022, opened 2022-07-20
Skip local tests for Client::getVersion()

thomascorthals:
@mkalkbrenner I found a way to pass the tests for `Client::getVersion()` locally. But I don't like it. I've still opened a PR for it because maybe it can lead to a better idea?

It uses `composer.json`. That takes it out of the realm of our "internal affairs" and "out into the open". That's not where we should be fiddling around for testing purposes. To make matters worse, there is no `scripts-dev` so it has to rely on an environment variable. And that variable has to be accessed with a slightly different syntax on Windows so you can't simply test with `[ $COMPOSER_DEV_MODE -eq 1 ]`.

Another thing I tried in `composer.json` felt more acceptable: using `autoload-dev` to load a mock for `Composer\InstalledVersions`. It doesn't seem possible to prepend it before the real vendor namespace though. The actual implementation is always discovered first.

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -19,7 +19,12 @@ class ClientTest extends TestCase
     {
         $installedVersion = InstalledVersions::getPrettyVersion('solarium/solarium');
 
-        if ('dev-master' === $installedVersion) {
+        // @see https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+        $semverRegex =
+            '/^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/'
+        ;
+
+        if (!preg_match($semverRegex, $installedVersion)) {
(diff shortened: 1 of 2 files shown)


mkalkbrenner:
If you're on a local branch or on dev-master, the version string will not be a semantic version string. I think it would better if we check for that in both failing tests and simply skip them if the version string is not "sematic".

tests/ClientTest.php
```diff
@@ -2,44 +2,61 @@
...
+     *
+     * @var string
+     */
+    protected static $versionTag = '76.5.4';
+
+    public static function setUpBeforeClass(): void
+    {
+        $installedVersion = InstalledVersions::getPrettyVersion('solarium/solarium');
+
+        if ('dev-master' === $installedVersion) {
```
mkalkbrenner: We should use ```strpos($installedVersion, 'dev-') === 0``` to cover feature branches, too.

tests/ClientTest.php
```diff
@@ -2,44 +2,61 @@
...
 use Solarium\Client;
 
 class ClientTest extends TestCase
 {
+    /**
+     * The version tag we use for testing within github actions.
+     *
+     * @var string
+     */
+    protected static $versionTag = '76.5.4';
```
mkalkbrenner, line 16: We currently have test failures for release tags, namely 6.2.6.
I think we should use ```InstalledVersions::getPrettyVersion('solarium/solarium')``` and use its value if it is a valid semantic version number.

mkalkbrenner (changes requested):

tests/ClientTest.php
thomascorthals: That was temporary to test local vs. github for which I had to push. ;-)

PR #1026, opened 2022-07-23
Proxy support for Http adapter

### LINKED ISSUE #861: Missing proxy-Support for Http-Adapter

thomascorthals:
Closes #861. Closes #862.

I needed an integration test with a proxy as an extra precaution for an upcoming PR that will change code in the adapters. Chose `ubuntu/squid` because it's actively maintained by Canonical on Docker Hub.

The lack of such a test was a hurdle for adding proxy support to the Http adapter. With the test in place, I took a shot at it.

@Radon8472 @wickedOne @dmaicher You were all involved in the previous PR. Your feedback is highly appreciated.

I went with a `ProxyAwareInterface` and `Trait` in the spirit of the timeouts. Both the `Curl` and `Http` adapters are fine with a _server:port_ string, but I don't assume that will be the case for every custom adapter that might be used. I deliberately don't enforce a type in the method signatures to allow for use cases where e.g. more complex settings are passed as an array.

The docs were augmented so there's no ambiguity about what each adapter accepts by default. I've also added a generic example of how to customize an existing adapter to add (any) additional options. This example isn't about proxies specifically, it's up to the user to connect the dots. I did implement additional integration tests in this way though, to prove that the documented principle actually works for this.

Finally, I deprecated setting a proxy through the options on the `Curl` adapter. Good illustration of what a pain in the endpoint `Configurable` can be. Without doing this explicitly now, we would never be able to remove this in the future without causing unexpected breakage for users.


dmaicher:
I still feel this adds a lot of code and complexity and it can already be achieved using PSR-18 and for example Guzzle or Symfony's http client. See https://github.com/solariumphp/solarium/pull/862#issuecomment-719909900.

thomascorthals:
@dmaicher If you follow Codecov's metric, this PR only has 21 lines of "actual" code. That includes the circuitous deprecations I added to the `Curl` adapter to wean it off `Configurable`. Personally I don't mind this (IMHO minor) complexity tradeoff as it doesn't even affect PSR-18 users.

The fact is that we can't abolish the `Curl` adapter because it's needed for parallel execution. I also think it's a good starter adapter for new users who just spun up their first techproducts core. The integration tests are a strong guarantee that you can switch to a PSR-18 adapter at any point. (The only caveat being parallel execution if you use a proxy, and that's already documented.)

I do agree that we don't need to add further complexity beyond this. We offer a solid base that works out of the box for the majority of use cases. You can extend the `Curl` adapter to add an extra option to a cURL transfer in 5 lines of code. Or you can switch to PSR-18. Whatever suits your comfort zone or the company policy about vetting external dependencies.

thomascorthals:
By the way, if you (that's a general you towards every interested party) think that setting a proxy through the options on a `Curl` adapter shouldn't be deprecated, feel free to make your case. It's literally a matter of removing two lines of code and a few `except`s in the unit tests to change this.

Radon8472:
I checked the changes, and it seems toto add the functionality that I tried to add with my PR.
And I agree that the "real" changes are not very complex. 

The pr only seems to be complex, because it adds proxy tests with docker too.
From my side, I think this changes would be very usefull, and the code looks good and fits the structure or the other code files in this project.

Issue #831, opened 2020-07-04
Solr Authentication


Firesphere:
With Solr now fully supporting several different types of authentication, it would be good to document how to set-up authentication for Solarium

mkalkbrenner:
Feel free to do it ;-)

Firesphere:
Heh, although I would agree, I've not fully figured it out yet, hence the issue instead of a PR :D

wickedOne:
by looking at the code, it looks like solarium currently only supports basic auth by adding a ``Authorization`` header in the request (have a look in the ``\Solarium\Core\Client\Adapter`` namespace).

kerebos, hadoop & jwt security plugins do not have a solarium implementation.

mkalkbrenner:
#948 introduced the usage of security.json in our tests. This one uses basic authentication.

mkalkbrenner:
I started experimenting with JWT and token based authentication in general.

mkalkbrenner:
I close this issue as authorization tokens are now supported.

If there's really a need for more authentications methods, open dedicated issues.

Firesphere:
Cheers @mkalkbrenner

PR #1031, opened 2022-09-09
added authorization token support

mkalkbrenner:


mkalkbrenner:
@thomascorthals It would be great if you review that PR. If you agree on the implementation, I'll add some documentation about authorization methods.

BTW unfortunately set Authentication() was introduced as function name for basic authentication. I suggest not to change that and to explicitly name the others that will be added. See #831

src/Core/Client/Adapter/Curl.php
```diff
@@ -107,6 +107,14 @@ public function createHandle(Request $request, Endpoint $endpoint)
             curl_setopt($handler, CURLOPT_USERPWD, $authData['username'].':'.$authData['password']);
             curl_setopt($handler, CURLOPT_HTTPAUTH, CURLAUTH_BASIC);
         }
+        elseif (!isset($options['headers']) || !array_key_exists('Authorization', $options['headers'])) {
```
thomascorthals, line 110:
```suggestion
        elseif (!isset($options['headers']) || !\array_key_exists('Authorization', $options['headers'])) {
```

In keeping with https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/866715801ac121a0a4cde6fdb9d9a3dd6873b606/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L347

src/Core/Client/Adapter/Curl.php
```diff
@@ -107,6 +107,14 @@ public function createHandle(Request $request, Endpoint $endpoint)
...
+            // According to the specification, only one Authorization header is allowed.
+            // @se https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
```
thomascorthals, line 112:
```suggestion
            // @see https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
```

src/Core/Client/Adapter/Http.php
```diff
@@ -103,6 +103,16 @@ public function createContext(Request $request, Endpoint $endpoint)
                 'Authorization: Basic '.base64_encode($authData['username'].':'.$authData['password'])
             );
         }
+        else {
+            // According to the specification, only one Authorization header is allowed.
+            // @se https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
```
thomascorthals, line 108:
```suggestion
            // @see https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
```

src/Core/Client/Adapter/Psr18Adapter.php
```diff
@@ -171,6 +171,14 @@ private function getRequestHeaders(Request $request, Endpoint $endpoint): array
             if (!empty($authData['username']) && !empty($authData['password'])) {
                 $headers['Authorization'] = ['Basic '.base64_encode($authData['username'].':'.$authData['password'])];
             }
+            else {
+                // According to the specification, only one Authorization header is allowed.
+                // @se https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
```
thomascorthals, line 176:
```suggestion
                // @see https://stackoverflow.com/questions/29282578/multiple-http-authorization-headers
```

thomascorthals (changes requested):
Looks good. Just some nitpicks on top of the ones from StyleCI.

And it appears you didn't commit an upated `CurlTest`?

mkalkbrenner:
@thomascorthals we don't have any authorization tests for curl so far. So we should also add something for basic auth.

thomascorthals (dismissed):

thomascorthals:
I assume I'm not the only one getting annoyed that there is no `curl_getopt()` for testing stuff like this.

PR #1032, opened 2022-09-12
Remove parent::init() if it calls Configurable::init()

thomascorthals:
There `parent::init()`s don't do anything except calling the empty method `Configurable::init()`.


thomascorthals:
Don't mind Codecov, it doesn't like covered code being deleted.

PR #1033, opened 2022-09-22
JSON requests for Update queries

thomascorthals:
This is my first shot at JSON Update requests. I didn't write any integration tests yet, hoping to get some early feedback first. Can be merged without integration tests IMHO as I kept XML requests as the default for now. I'll add the integration tests later on anyway. Documentation will also follow when I'm sure the implementation is stable.

The choice "between XML and JSON" is implemented as a choice "for a request format" to allow adding other formats in the future. (If anyone ever needs to send CSV requests …)

It's a bit unfortunate that building the raw JSON data still requires string concatenation in the code. There is no data structure in PHP that can be `json_encode()`d directly to a JSON object with non-unique keys.

I would keep JSON requests opt-in for the first release. Once released, I can use them as an early adopter on a platform that generates 5 to 6 requests/second, including a lot of atomic updates. I don't use nested documents, maybe that's something @jupevi can test as an early adopter?

I made some choices w.r.t. backward compatibility in favour of reducing code complexity.

1. Boosts are ignored completely because they're no longer supported at index-time since Solr 7 and there is no JSON syntax for them anyway. Using boosts with JSON request could be made to throw a RuntimeException, but that's a pointless performance hit. A user actively switching to JSON requests should already be aware that they don't support boosts.
2. A `RawXml` command throws the generic exception for an unsupported command. It could be made to throw a specific message instructing you to use XML requests, but it should be obvious that you can't mix in XML with JSON requests.

I've kept the unit tests for XML and JSON request builders in sync, even when certain tests are specifically aimed at an implementation detail of one of the formats (e.g. escaping `<` and `>` for XML or multivalued DateTimes for JSON).

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -14,7 +14,7 @@ jobs:
 
         strategy:
             matrix:
-                php: [7.3, 7.4, 8.0, 8.1]
+                php: [7.3, 7.4, 8.0, 8.1, 8.2]
                 solr: [7, 8, 9]
                 mode: [cloud, server]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- PHP 8.2 support
 - JSON formatted update requests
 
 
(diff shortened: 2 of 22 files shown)


mkalkbrenner:
@thomascorthals thanks for providing that implementation. I agree to merge it as it is and extend the test coverage later one the merge conflict is solved.

PR #1035, opened 2022-10-09
Test that Helper::formatDate() doesn't modify input

### LINKED ISSUE #840: Helper::formatDate

thomascorthals:
@wickedOne @mkalkbrenner I think we can close #840 and close #841.

#613 already fixed this:

https://github.com/solariumphp/solarium/blob/2b063e971c1ecddc2a43a4bc6c35fa0ffb0eaf3a/src/Core/Query/Helper.php#L111-L114

The unit test from that PR is easy to miss though because it's not in `HelperTest`. I've moved it and rewritten it to be more "readable" as to what is being tested. Also switched to offset values when testing with timezones because that makes the tests easier to verify (without e.g. having to know about a location's DST rules). Testing 'Europe/London' against 'UTC' is a bad idea anyway, that could mask a hypothetical bug where only the timezone indicator is changed without adjusting the datetime accordingly.

And I changed the date formatting to a one-liner that doesn't require further string manipulation.

As far as the function name is concerned, I think `formatDate()` is perfectly acceptable if you consider its FQN `Solarium\Core\Query\Helper::formatDate()`. It's a method that _helps_ you _format a date_ to use in a _query_, and there's only one format that Solr understands. If you want to do things with your date outside of queries, this is not the namespace you're looking for. _jedi hand wave_

### DIFF
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -142,7 +142,7 @@ class Helper
     public function formatDate($input)
     {
         switch (true) {
-            // input of datetime object
+            // input of DateTime or DateTimeImmutable object
             case $input instanceof \DateTimeInterface:
                 $input = clone $input;
                 break;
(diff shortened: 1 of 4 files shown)


src/Core/Query/Helper.php
```diff
@@ -172,15 +172,10 @@ public function formatDate($input)
 
         // handle the filtered input
         if ($input) {
-            // when we get here the input is always a datetime object
+            // when we get here $input is always a DateTime or DateTimeImmutable object
```
thomascorthals, line 175: I know this is nitpicking, but `$input` when we get here is no longer identical to "the input" that was passed as a parameter to the method.

PR #1042, opened 2022-11-19
PHPDoc fixes

thomascorthals:
Found a few easily fixable issues with phpstan at level 2.

(Was wondering if we could go to level 2 overall. Doesn't look feasible right now.)


thomascorthals:
Codecov thinks that I removed coverage, but I didn't.

PR #1046, opened 2022-12-07
For Solarium 7: set default to JSON formatted update requests

thomascorthals:
@mkalkbrenner I'm in no rush for Solarium 7. Just wanted to have this PR ready while I still had a fresh recollection of what needed changing in the docs.


mkalkbrenner:
@thomascorthals I think we should go for 7.0.0-beta.1 as soon as you report that it works in your environment.

PR #1048, opened 2022-12-10
JSON serialization of arrays with non-consecutive indices in multivalue fields

thomascorthals:
Ran into an issue with JSON requests with real-life data. It always worked with XML requests so it is or does at least feel like a BC break. PRed against the `6.x` branch because that will need a bugfix release. Will have to go into `master` as well. (If you agree it's something to fix in the library and not leave up to the users.)

Our userland code uses `array_unique()` on some arrays that go into multivalue fields and that results in non-consecutive array indices (unlike some other PHP functions which renumber numeric keys). Same thing could happen with `unset()`. This doesn't matter for XML because the request builder `foreach`es over the values and doesn't even look at the keys. It does matter for JSON because `json_encode()` turns them into objects instead of arrays.

```php
echo json_encode([0, 1, 2]);
// [0,1,2]

echo json_encode(array_unique([0, 1, 1, 2]));
// {"0":0,"1":1,"3":2}
```

This leads to a SolrException:

```json
{
  "responseHeader":{
    "status":400,
    "QTime":0},
  "error":{
    "metadata":[
    "msg":"Unknown operation for the an atomic update, operation ignored: 0 for id:SP2514N",
    "code":400}}
```

Unfortunately, there is no flag for `json_encode()` to force it to serialize this as an array regardless. That leaves us with running it through `array_values()` first.

I feel that as a library, that is something that we should do. There is no reason to doubt the user's intent here, and the XML request builder de facto assumed this was the intended behaviour (probably without thinking about it that way, it just happened to work out).

The other option would be to leave it up to the user to ensure array indices are consecutive. Then we'd need to get that point across in the documentation. As a user, I'd rather have the library worrying about such implementation details. Especially since you can luck out 99 times with `array_unique()`/`unset()` happening to delete from the end and only encounter it the 100th time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### Added
+### Fixed
+- JSON serialization of arrays with non-consecutive indices in multivalue fields
 
 
 ## [6.2.8]
--- a/src/QueryType/Update/Query/Document.php
+++ b/src/QueryType/Update/Query/Document.php
@@ -499,6 +499,9 @@ class Document extends AbstractDocument
             if ($value instanceof \DateTimeInterface) {
                 $value = $this->getHelper()->formatDate($value);
             } elseif (\is_array($value) && is_numeric(array_key_first($value))) {
+                // ensure consecutive indices so it doesn't serialize to an object
+                $value = array_values($value);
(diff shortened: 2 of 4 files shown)


mkalkbrenner:
I run Drupal's search_api_solr tests using solarium master and run into the same issue. So yes, we should handle it in the library!

mkalkbrenner:
Actually it wasn't the same error but the legacy tests with Solr 3.6 failed. I know that it isn't officially supported by solarium anymore, but that legacy integration just uses a minimal subset of the features. I need to change its code to force XML. The newer Solr versions will use JSON in the tests :-)

PR #1053, opened 2023-02-16
Trailing question mark when query string is empty

drunken-monkey:
Currently, a request without GET parameters will still include a trailing `?` character in the URL. This commonly occurs when using the `PostBigRequest` plugin, and can lead to unnecessary problems in combination with a server that forbids the mixing of GET and POST parameters.

While it is clear that, normally, a web server should be expected to treat an empty query string the same as a completely missing query (that is, the trailing `?` shouldn’t matter), I think we can also agree that the URL looks cleaner without the unnecessary trailing `? ` character. Furthermore, looking at the [URI specification of the W3C](https://www.w3.org/Addressing/URL/uri-spec.html), it seems a trailing `?` would even violate the standard, strictly speaking.

Unless there are good reasons to keep the trailing `?`, I propose the simple change of removing the `?` for empty query strings.

### DIFF
--- a/src/Core/Client/Request.php
+++ b/src/Core/Client/Request.php
@@ -425,7 +425,7 @@ class Request extends Configurable implements RequestParamsInterface
     /**
      * Get an URI for this request.
      *
-     * @return string|null
+     * @return string
      */
     public function getUri(): string
     {
--- a/src/Core/Client/Request.php
+++ b/src/Core/Client/Request.php
@@ -427,11 +427,11 @@ class Request extends Configurable implements RequestParamsInterface
      *
      * @return string|null
      */
-    public function getUri(): ?string
+    public function getUri(): string
     {
         $queryString = $this->getQueryString();
         if ('' === $queryString) {
-            return $this->getHandler();
+            return $this->getHandler() ?? '';
         }
 
         return $this->getHandler().'?'.$queryString;
--- a/src/Core/Client/Request.php
+++ b/src/Core/Client/Request.php
@@ -429,7 +429,12 @@ class Request extends Configurable implements RequestParamsInterface
      */
     public function getUri(): ?string
     {
-        return $this->getHandler().'?'.$this->getQueryString();
+        $queryString = $this->getQueryString();
+        if ('' === $queryString) {
+            return $this->getHandler();
+        }
+
+        return $this->getHandler().'?'.$queryString;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


thomascorthals:
Not going to argue with Tim Berners-Lee of course. 😉 But that URI specification is from 1994 so I did have a look at more recent sources.

Having a trailing `?` doesn't violate [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986) which defines these ABNF representations for `URI` and `query`.

> ```
> URI           = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
> 
> query         = *( pchar / "/" / "?" )
> ```

However, it does violate the serialisation rules in WHATWG's [URL Standard](https://url.spec.whatwg.org/#url-serializing).

> 5. If _url_’s [query](https://url.spec.whatwg.org/#concept-url-query) is non-null, append U+003F (?), followed by _url_’s [query](https://url.spec.whatwg.org/#concept-url-query), to _output_.

I do agree that it's better to omit it in this case.

src/Core/Client/Request.php
```diff
@@ -429,7 +429,12 @@ public function clearHeaders(): self
...
     public function getUri(): ?string
     {
-        return $this->getHandler().'?'.$this->getQueryString();
+        $queryString = $this->getQueryString();
+        if ('' === $queryString) {
+            return $this->getHandler();
+        }
+
+        return $this->getHandler().'?'.$queryString;
     }
```
thomascorthals, line 438: @mkalkbrenner I'm not entirely sure why we would ever want to return `null` here? The return value is always used as a string.

```php
    public function getUri(): string
    {
        $handler = $this->getHandler() ?? '';
        $queryString = $this->getQueryString();

        if ('' === $queryString) {
            return $handler;
        }

        return $handler.'?'.$queryString;
    }
```

mkalkbrenner: The definitions of return values have been added as a mass update. We already discovered wrong types from time to time. I think that string is correct here.

thomascorthals (changes requested):
@drunken-monkey There's nothing with your PR, but do you mind changing the return type to `string` instead of `?string` and updating the unit test accordingly while you're at it? 😇

drunken-monkey:
@thomascorthals Sure, no problem.

src/Core/Client/Request.php
```diff
@@ -427,9 +427,14 @@ public function clearHeaders(): self
      *
      * @return string|null
```
thomascorthals, line 428:
```suggestion
     * @return string
```

thomascorthals (approved):

PR #1057, opened 2023-03-18
StyleCI fixes

thomascorthals:
Ran into some subtle StyleCI changes unrelated to what I was working on.

### DIFF
--- a/src/Core/Query/Helper.php
+++ b/src/Core/Query/Helper.php
@@ -142,13 +142,13 @@ class Helper
     public function formatDate($input)
     {
         switch (true) {
-            // input of DateTime or DateTimeImmutable object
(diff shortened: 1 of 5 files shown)


src/Core/Query/Helper.php
```diff
@@ -161,11 +161,8 @@ public function formatDate($input)
                     $input = false;
                 }
                 break;
-            // any other input formats can be added in additional cases here...
-            // case $input instanceof Zend_Date:
```
thomascorthals, line 165: StyleCI wanted to indent this one more stop. Looks off to have `// case` line up with `break;`. It's safe to assume that contributors know how to add a case without these instructions. 😉

PR #1065, opened 2023-05-03
Fix PHP 8.2 deprecations

### LINKED ISSUE #1063: PHP 8.2: Deprecated notices in Core/Query/Result/QueryType.php
**Solarium version(s) affected**: 6.2.8

**Description**  
Following deprecated notices:

PHP-ERROR.NOTICE: Creation of dynamic property Solarium\QueryType\Server\CoreAdmin\Result\Result::$responseHeader is deprecated - Core/Query/Result/QueryType.php line 56

PHP-ERROR.NOTICE: Creation of dynamic property Solarium\QueryType\Server\CoreAdmin\Result\Result::$initFailures is deprecated - Core/Query/Result/QueryType.php line 56

PHP-ERROR.NOTICE: Creation of dynamic property Solarium\QueryType\Server\CoreAdmin\Result\Result::$status is deprecated - Core/Query/Result/QueryType.php line 56

PHP-ERROR.NOTICE: Creation of dynamic property Solarium\QueryType\Server\CoreAdmin\Result\Result::$queryTime is deprecated - Core/Query/Result/QueryType.php line 56

**How to reproduce**  
```
$query = $solr->createCoreAdmin();
$query->setAction( $query->createStatus()->setCore( 'core' ) );
$result = $solr->coreAdmin( $query )->getStatusResult();
```

thomascorthals:
Fixes #1063.

Any key that can be present in Solr's response data must be declared as a property of the result class it gets parsed into to avoid 'creation of dynamic property' deprecations.

I've moved the handling of the `responseHeader`'s `status` and `QTime` to the parent `QueryType` result to make it consistent across all QueryTypes and not something individual response parsers have to bother with. Renamed three `getStatus()` methods with a name clash in the Collections results, with a deprecation so it doesn't break BC.

@mkalkbrenner The Collections and CoreAdmin results don't have any unit test coverage so I didn't include any for my additions either. Can we overlook the <100% patch coverage for once? 😉

### DIFF
--- a/src/Core/Query/AbstractResponseParser.php
+++ b/src/Core/Query/AbstractResponseParser.php
@@ -51,30 +51,6 @@ abstract class AbstractResponseParser
     }
 
     /**
-     * Parses header data (if available) and adds it to result data.
-     *
-     * @param array $data
-     * @param array $result
-     *
-     * @return array
-     */
-    public function addHeaderInfo(array $data, array $result): array
-    {
-        $status = null;
-        $queryTime = null;
-
-        if (isset($data['responseHeader'])) {
-            $status = $data['responseHeader']['status'];
-            $queryTime = $data['responseHeader']['QTime'];
-        }
-
-        $result['status'] = $status;
-        $result['queryTime'] = $queryTime;
-
-        return $result;
-    }
-
-    /**
      * Parses HTTP status code and adds boolean wasSuccessful to result data.
      * Parses HTTP status message and adds string statusMessage to result data.
      *
--- a/src/QueryType/Ping/Query.php
+++ b/src/QueryType/Ping/Query.php
@@ -31,7 +31,7 @@ class Query extends BaseQuery
     protected $options = [
         'resultclass' => Result::class,
         'handler' => 'admin/ping',
-        'omitheader' => true,
+        'omitheader' => false,
     ];
 
     /**
@@ -55,12 +55,12 @@ class Query extends BaseQuery
     }
 
     /**
-     * The ping query has no response parser so we return a null value.
+     * Get a responseparser for this query.
      *
-     * @return \Solarium\Core\Query\ResponseParserInterface|null
+     * @return ResponseParser
      */
     public function getResponseParser(): ?ResponseParserInterface
     {
-        return null;
+        return new ResponseParser();
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - JSON serialization of arrays with non-consecutive indices in multivalue fields
+- PHP 8.2 deprecations
 
 ### Changed
 - Update queries use the JSON request format by default
+
+### Deprecated
+- Solarium\QueryType\Server\Collections\Result\CreateResult::getStatus(), use getCreateStatus() instead
+- Solarium\QueryType\Server\Collections\Result\DeleteResult::getStatus(), use getDeleteStatus() instead
+- Solarium\QueryType\Server\Collections\Result\ReloadResult::getStatus(), use getReloadStatus() instead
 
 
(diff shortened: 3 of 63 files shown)


thomascorthals:
@mkalkbrenner If the StyleCI failure is blocking, #1057 will have to be merged first.

src/QueryType/Ping/Query.php
```diff
@@ -31,7 +31,7 @@ class Query extends BaseQuery
     protected $options = [
         'resultclass' => Result::class,
         'handler' => 'admin/ping',
-        'omitheader' => true,
+        'omitheader' => false,
```
mkalkbrenner, line 34: Could you explain that change?

src/Core/Query/AbstractResponseParser.php
```diff
@@ -50,30 +50,6 @@ public function convertToKeyValueArray(array $data): array
...
 
-    /**
-     * Parses header data (if available) and adds it to result data.
-     *
-     * @param array $data
-     * @param array $result
-     *
-     * @return array
-     */
-    public function addHeaderInfo(array $data, array $result): array
```
mkalkbrenner, line 61: I wonder if that change breaks backward compatibility. If I read the PR correctly, status and queryTime have to be accessed differently now by the client.

src/QueryType/Ping/Query.php
thomascorthals: Ping result used to have this hardcoded:

```php
    public function getStatus(): int
    {
        return 0;
    }
```

With my changes, calling `$result->getStatus()` on a ping calls the generic implementation in `Core\Query\Result\QueryType`. To maintain BC and still have it return `0` on success, the responseHeader must be included in the response.

src/Core/Query/AbstractResponseParser.php
thomascorthals: You can still call `->getStatus()` and `->getQueryTime()` on the results that already had those methods. With my changes, you can now call it for every query type's result, not just the ones it happened to be implemented for.

AFAIK it doesn't break BC. The changes I had to make to the unit tests were only for how a result stores this info now, and those are `protected` members so that doesn't matter to the client. If you do find a way to break BC, let me know of course!

thomascorthals: One way to check this is `examples/1.3-basic-update.php`. It still shows the status and query time without modifications.

thomascorthals: @mkalkbrenner I went over it again and I get your concern now. I was solely focused on status and queryTime so I'm confident I didn't break those.

Now I'm wondering if we use other parameters from the responseHeader. If I broke something there, it wouldn't have had test coverage though or I would have been aware. I'll convert this PR to a draft until I've had the time to doublecheck this to make sure.

mkalkbrenner: In Drupal we have a debug mode for solarium queries. I should check if it still works.

thomascorthals: Turns out I did break something: if a reponse parser "cherry picked" from the response data, it ignored the response header. This wasn't caught in any of the unit/integration tests and of course I happened to only use query types where this wasn't an issue for my manual verification 🙄. Fixed that and added a unit test.

I also sifted through the Solr source code for every parameter that can be added to `responseHeader`. This is the complete list:
- status
- QTime
- handler
- params
- partialResults
- segmentTerminatedEarly
- zkConnected
- warnings
- rf
- msg
- errors
- maxErrors

Of these, only `status` and `QTime` are always set, and those are also the only ones that were captured by `addHeaderInfo()`. Others depend on handler and some on cloud mode too. `Ping\Result::getZkConnected()` demonstrates how it's trivial to add support for those.

src/QueryType/Ping/Query.php
thomascorthals: I included this in the CHANGELOG and mentioned it in the docs.

PR #1091, opened 2023-08-04
Add min/max scaling to query re-ranking

thomascorthals:
[`reRankScale`/`reRankMainScale`](https://solr.apache.org/guide/solr/latest/query-guide/query-re-ranking.html#rerank-query-parser) was added in [SOLR-16827](https://issues.apache.org/jira/projects/SOLR/issues/SOLR-16827) for Solr 9.3.


thomascorthals:
I have considered `public function setScale(int $min, int $max)` instead of validating the `min-max` format but that didn't mesh well with `getScale()` and config mode.

Issue #1095, opened 2023-09-26
Spatial component should use floats for distance


wu-edward:
**Solarium version(s) affected**: 6.3.1
**Solr version**: 8.11.2
**Solr mode**: standalone

**Description**
The spatial component only allows integer values for distance. This can cause a lack of precision if there is a need to convert between units (such as mi -> km). See https://www.drupal.org/project/search_api_solr/issues/3389956 for more details.

**How to reproduce**

**Possible Solution**
Change \Solarium\Component\Spatial getDistance, and setDistance methods to set and return distance as float

**Additional context**

thomascorthals:
Good catch. The parameter is indeed parsed as a `double` in Solr.

https://github.com/apache/solr/blob/4c80cc2ef9e50915fb7aaa6e9cfe042a27d2c44d/solr/core/src/java/org/apache/solr/search/SpatialFilterQParser.java#L59-L61

```java
    double dist = -1;
    String distS = getParam(SpatialParams.DISTANCE);
    if (distS != null) dist = Double.parseDouble(distS);
```

thomascorthals:
Fixed in #1096.

PR #1096, opened 2023-09-26
Issue #1095: Spatial component should use floats for distance.

wu-edward:


wu-edward:
Addresses #1095

thomascorthals:
Minor nitpick: maybe update these tests to use `float`s? That should prevent the unlikely regression of accidentally reverting to `int`s.

https://github.com/solariumphp/solarium/blob/376ad9fdd56cb07f76a3b7c1dff6608b143d5c18/tests/Component/SpatialTest.php#L21

https://github.com/solariumphp/solarium/blob/376ad9fdd56cb07f76a3b7c1dff6608b143d5c18/tests/Component/SpatialTest.php#L65

https://github.com/solariumphp/solarium/blob/376ad9fdd56cb07f76a3b7c1dff6608b143d5c18/tests/Component/RequestBuilder/SpatialTest.php#L12

wu-edward:
> Minor nitpick: maybe update these tests to use `float`s? That should prevent the unlikely regression of accidentally reverting to `int`s.
> 
> https://github.com/solariumphp/solarium/blob/376ad9fdd56cb07f76a3b7c1dff6608b143d5c18/tests/Component/SpatialTest.php#L21
> 
> https://github.com/solariumphp/solarium/blob/376ad9fdd56cb07f76a3b7c1dff6608b143d5c18/tests/Component/SpatialTest.php#L65
> 
> https://github.com/solariumphp/solarium/blob/376ad9fdd56cb07f76a3b7c1dff6608b143d5c18/tests/Component/RequestBuilder/SpatialTest.php#L12

Thanks for the review. Pushed a commit with updated tests.

thomascorthals (approved):

thomascorthals:
@wu-edward I often forget this myself for small fixes like this, but it's worth mentioning in the `CHANGELOG` too. No longer rounding to an integer distance will influence results.

wu-edward:
> @wu-edward I often forget this myself for small fixes like this, but it's worth mentioning in the `CHANGELOG` too. No longer rounding to an integer distance will influence results.

CHANGELOG updated

Issue #1104, opened 2023-11-20
EscapePhrase with \r does not work 


alexander-schranz:
**Solarium version(s) affected**: 6.x-dev b785b8b
**Solr version**: 9
**Solr mode**: cloud

**Description**

In my search engine abstraction I try to filter by a very special field:

```php
$specialString = "^The 17\" O'Conner && O`Series \n OR a || 1%2 \r\n book? \r \twhat \\ text // ok? end$";
```

Why most of the special chars work with the `escapePhrase` method it does not work for the `\r` value. So I thin kthere is something missing.

**How to reproduce**

Index a document with that special chars.

Try to filter by it: 

```php
'tag.raw=' . $helper->escapePhrase($specialString); 
```

**Possible Solution**

Not sure what is missing in `escapePhrase`.

**Additional context**

```
// does not work:
$specialString = "^The 17\" O'Conner && O`Series \n OR a || 1%2 \r\n book? \r \twhat \\ text // ok? end$";
// works:
$specialString = "^The 17\" O'Conner && O`Series \n OR a || 1%2 \n book? \twhat \\ text // ok? end$";
```

thomascorthals:
I'm not sure what's happening here. Only `"` and `\` are escaped by `escapePhrase()` so it shouldn't treat `"\r\n"` any differently than `"\n"`.

Did you index through Solarium as well? If so, can you provide a code sample to reproduce this behaviour?

Either way, can you verify that the `\r` value is actually present in the index? If by "works" you mean you can retrieve the document without `\r` in the query, I suspect it never made it into the field in the first place.

alexander-schranz:
Yeah it is indexed also via Solarium. You can checkout the failing test here: https://github.com/schranz-search/schranz-search/pull/302

Checkout the PR:

```
git clone git@github.com:alexander-schranz/schranz-search.git
cd schranz-search
git checkout bugfix/solr-filter
cd packages/seal-solr-adapter/

docker compose up # takes some time until solr is booted

# run the test

composer test -- --filter="testEqualConditionSpecialString"
```

I will check it at the evening if the `\r` is still there when read from the index.

alexander-schranz:
Solr itself does not strip the `\r` away reading and outputing the docs the `\r` still is there:

> `echo (json_encode($loadedDocuments[0]['tags'][2]));`
> "^The 17\" O'Conner && O`Series \n OR a || 1%2 \r\n book? \r \twhat \\ text \/\/ ok? end$"

[...]

PR #1106, opened 2023-11-21
Add "\n", "\t", "\r" to test case

alexander-schranz:
Add some more test cases: https://github.com/solariumphp/solarium/issues/1104


thomascorthals:
#1107 has a working test that incorporates the test phrase from #1104 and accounts for XML's handling of `\r`.

PR #1119, opened 2024-01-03
support suggest.buildAll and add a plugin to execute queries without waiting for Solr's response

mkalkbrenner:
We noticed that building suggesters can be very expensive.
PHP scripts or their HTTP connections to Solr might run into timeouts on large indexes.
The NoWaitForResponseRequest plugin introduces a kind of fire-and-forget queries.


mkalkbrenner:
@thomascorthals documentation needs to be added if accepted

thomascorthals (changes requested):
Not sure about the use of `TimeoutAwareInterface::MINIMUM_TIMEOUT`.

From <https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html>:

> With [CURLOPT_CONNECTTIMEOUT](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html) set to 4 and [CURLOPT_TIMEOUT](https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html) set to 2, the operation can never last longer than 2 seconds.

If the adapter is also ConnectionTimeoutAware, the plugin should add that timeout to `MINIMUM_TIMEOUT` I think.

src/Core/Client/Adapter/TimeoutAwareInterface.php
```diff
@@ -21,6 +21,11 @@ interface TimeoutAwareInterface
      */
     public const DEFAULT_TIMEOUT = 5;
 
+    /**
+     * Minimum timeout that should be respected by adapters implementing this interface.
```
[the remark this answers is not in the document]
mkalkbrenner: I changed it to FAST_TIMEOUT

mkalkbrenner:
@thomascorthals thanks for your review. I modified the code and added some minimal documentation.

thomascorthals:
It still doesn't take the connection timeout into account in the plugin. I wrote this example (I'll commit the cleaned up version afterwards).

```php
<?php

require_once __DIR__.'/init.php';
htmlHeader();

// set a long connection timeout
$adapter->setConnectionTimeout(10);

// connect to a host that will never complete the connection
$config['endpoint']['localhost']['host'] = '10.0.0.0';

// create a client instance
$client = new Solarium\Client($adapter, $eventDispatcher, $config);

// get a suggester query instance and add setting to build all suggesters
$suggester = $client->createSuggester();
$suggester->setBuildAll(true);

// don't wait unitl all suggesters have been built
$client->getPlugin('nowaitforresponserequest');

// this executes the query
$client->suggester($suggester);

htmlFooter();
```

When I comment out the plugin, this results in:
> Solr HTTP error: HTTP request failed, Connection timeout after 10005 ms

With the plugin, this script ends after 1 second. If I `echo $e->getMessage()` in the plugin's `catch` block, I get:
> Solr HTTP error: HTTP request failed, Connection timed out after 1004 milliseconds

That should also be slightly above 10000 milliseconds.

And if I run it on `localhost` against `ncat -k -l 8983` (a listener that never responds), that should actually be slightly above 11000 milliseconds.

src/Plugin/NoWaitForResponseRequest.php
```diff
@@ -0,0 +1,123 @@
...
+            $request->setRawData($queryString);
+            $request->clearParams();
+        }
+
+        $timeout = TimeoutAwareInterface::DEFAULT_TIMEOUT;
+        if ($this->client->getAdapter() instanceof TimeoutAwareInterface) {
+            $timeout = $this->client->getAdapter()->getTimeout();
+            $fastTimeout = TimeoutAwareInterface::FAST_TIMEOUT;
+            if ($this->client->getAdapter() instanceof ConnectionTimeoutAwareInterface) {
+                $fastTimeout += $this->client->getAdapter()->getConnectionTimeout() ?? 0;
```
thomascorthals, line 66: This hinges on the assumption that all ConnectionTimeoutAware adapters will behave like Curl in this regard.

thomascorthals:
@mkalkbrenner Actually, we only want to return a `HTTP/1.0 200 OK` response for an operation timeout! A connection timeout should still throw an exception because in that case the request never went through and there is no way of telling with the current plugin implementation.

mkalkbrenner:
> @mkalkbrenner Actually, we only want to return a `HTTP/1.0 200 OK` response for an operation timeout! A connection timeout should still throw an exception because in that case the request never went through and there is no way of telling with the current plugin implementation.

@thomascorthals I just implemented that right before you posted this comment :-)
But I had to resolve the conflicts first.

src/Plugin/NoWaitForResponseRequest.php
```diff
@@ -0,0 +1,139 @@
...
+            } else {
+                $this->client->getAdapter()->setTimeout(TimeoutAwareInterface::FAST_TIMEOUT);
+            }
+        }
+
+        if ($this->client->getAdapter() instanceof Curl) {
+            $this->client->getAdapter()->setOption('return_transfer', false);
+        }
+
+        $microtime1 = microtime();
```
thomascorthals, line 77: I think this should be `microtime(true)`.

E.g. https://onlinephp.io?s=s7EvyCjg5eLlKkssik8pzS3QUCnJzE01VLBVyM1MLsoHcTQ0Na2RFaRhqigpKk0Fq-LlKs5JTS3QMIFw4JqQDFPQVYDYgWIqmlkgRWlQVQA%2C&v=8.2.13

```php
<?php

var_dump($time1 = microtime());
var_dump($ftime1 = microtime(true));

sleep(4);

var_dump(microtime() - $time1);
var_dump(microtime(true) - $ftime1);
```

```
string(21) "0.33706600 1704466431"
float(1704466431.337109)

Warning: A non-numeric value encountered in /home/user/scripts/code.php on line 8

Warning: A non-numeric value encountered in /home/user/scripts/code.php on line 8
float(0.00018300000000004424)
float(4.000219821929932)
```

mkalkbrenner:
@thomascorthals I assume it is finished now.

mkalkbrenner:
Now the failing tests proven that exceptions ar thrown ;-)

mkalkbrenner:
@thomascorthals Now that the test are passing, should we merge this one and got for a release to fix the documentation?

thomascorthals:
> @thomascorthals Now that the test are passing, should we merge this one and got for a release to fix the documentation?

@mkalkbrenner Go for it!

thomascorthals (dismissed):

PR #1134, opened 2024-10-09
PHP 8.4 support

thomascorthals:
https://github.com/php/php-src/blob/cc09e3f8d0030517ba4f355d8a576433549aab23/UPGRADING#L497-L498:

>   . Implicitly nullable parameter types are now deprecated.
>     RFC: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -14,7 +14,7 @@ jobs:
 
         strategy:
             matrix:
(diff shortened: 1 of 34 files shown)


thomascorthals:
@mkalkbrenner Pull requests are still convinced the PHP 7.3 and 7.4 checks are required. Can we change that?

PR #1137, opened 2024-11-04
Remove dead link from ExceptionInterface.php

kaznovac:
the url is last seen 2019-04-01 as per wayback machine while article is quite good and informative there's a little benefit of keeping it's reference in today's code doc

https://web.archive.org/web/20190401130536/http://ralphschindler.com/2010/09/15/exception-best-practices-in-php-5-3


thomascorthals:
All relevant information has since been included in the [documentation](https://solarium.readthedocs.io/en/stable/exceptions/) and [examples](https://github.com/solariumphp/solarium/blob/master/examples/6.5-exceptions.php).

PR #1144, opened 2025-01-20
PHPUnit 9 → 10

thomascorthals:
Major changes affecting us:

- `setUpBeforeClass()`/`tearDownAfterClass()` can no longer perform assertions. Tests of correct configuration have been moved to separate test functions.
- Changes in mocks necessitate a different approach, a.o. when testing traits.
- Data providers always have to be declared `static`.


thomascorthals:
Looks like I broke the codecov integration. Converting to draft for now.

thomascorthals:
Codecov worked after I let it access my branch. But it reported a drop in coverage. Not entirely sure where that came from, I filled in some unit tests for code that was only indirectly covered and it stopped complaining about that.

By then Solr 9.8 was released and that broke some integration tests. Things I had to chance to fix that:
- The `extraction` module is no longer loaded by default, fixed by adding the `SOLR_MODULES=extraction` environment variable.
- The patch file for `solrconfig.xml` is no longer compatible with previous versions, fixed by having a separate patch per version like we do for the schema.
- 1 of the 32 sample documents in `techproducts` was removed, re-added the minimally required fields for the integration tests rather than checking the version in every integration test to determine if we must expect 32 or 31 docs.
- For some reason the `solr1` instance in the cloud test didn't start unless I added `-z` (`--zk-host`) to the command even though the other containers have no problem recognising `ZK_HOST`.

```
solr1  | Neither --zk-host or --solr-url parameters provided so assuming solr url is http://localhost:8983.
solr1  |
solr1  | ERROR: Server refused connection at: http://localhost:8983/solr/admin/info/system
solr1  |
solr1 exited with code 1
```

thomascorthals:
The Solr 9 tests fail since Solr 9.8 was released. They will pass when #1146 is merged.

mkalkbrenner (approved):

PR #1151, opened 2025-04-04
Style fixes

thomascorthals:
Was working on something else when StyleCI started complaining about two files I hadn't touched. Looks like they've become a bit stricter with `use` statements.


thomascorthals:
@mkalkbrenner You'll need to configure an upload token for Codecov if we want to keep checking code coverage on the repo.

PR #1160, opened 2025-07-30
Update `Result::jsonSerializable` return type

ro0NL:
See https://github.com/symfony/symfony/issues/61275

I choose `array` as a more specific type, but im also fine with `mixed` from the contract.

### DIFF
--- a/src/Core/Query/Result/Result.php
+++ b/src/Core/Query/Result/Result.php
@@ -131,6 +131,8 @@ class Result implements ResultInterface, \JsonSerializable
     #[\ReturnTypeWillChange]
     /**
      * {@inheritdoc}
+     *
+     * @return array
      */
     public function jsonSerialize()
     {


thomascorthals:
Might as well do it for every class that has a `jsonSerialize()`.

ro0NL:
Perhaps, i was just following our logs for now.

I can test a bit longer, but i wasnt hoping to do the full maintenance at once here now.

thomascorthals:
> But if master is 7.x, we'll do that upgrade.

Next release will most likely still be 6.x. It's just that:

https://github.com/solariumphp/solarium/blob/894cedfd9e957fa30b225bdd0333ee9d1b7bebfb/CONTRIBUTING.md?plain=1#L20

And the workflow in the branch you targeted no longer works so the tests didn't run on that. :)

thomascorthals:
Looks like #1161 also fixes the deprecation. That being said, I'm not opposed to declaring explicit return types. I think we can actually remove `#[ReturnTypeWillChange]` if we do? That was the solution when we still supported PHP 7 that didn't have a `mixed` type.

PR #1168, opened 2025-11-16
Typed properties

thomascorthals:
A new major version is the perfect opportunity to do some overhaul on the codebase. I've replaced the type hints with actual type declarations on class properties (introduced in PHP 7.4). And I've added union type declarations to method signatures (introduced in PHP 8.0) where this was previously also only done in type hints.

This is a breaking change for anyone extending our codebase, which I've added to the pitfalls. But it shouldn't break much for regular use of the library since I didn't have to change anything in the integration tests and they still pass.


mkalkbrenner:
@thomascorthals I agree, we should do that for 7.0.0.
But what do you think? Shouldn't we declare interfaces for every class and use these interfaces as type properties?

thomascorthals:
> But what do you think? Shouldn't we declare interfaces for every class and use these interfaces as type properties?

Hadn't really thought about that. Would add additional work on our side to keep them in sync.

I was actually looking into getting the codebase to PHPStan level 2 and a lot of the errors seem to be caused by type hinting with interfaces that don't define all methods that are present in the implementing class. Here's an example, there are currently dozens like this.

```
 ------ ---------------------------------------------------------------------------------------------
  Line   Component\Facet\JsonFacetTrait.php (in context of class Solarium\Component\Facet\JsonTerms)
 ------ ---------------------------------------------------------------------------------------------
  140    Call to an undefined method Solarium\Component\Facet\FacetInterface::serialize().
         🪪  method.notFound
 ------ ---------------------------------------------------------------------------------------------

 ------ -----------------------------------------------------------------------------------
  Line   Component\RequestBuilder\FacetSet.php
 ------ -----------------------------------------------------------------------------------
  55     Call to an undefined method Solarium\Component\Facet\FacetInterface::getField().
         🪪  method.notFound
  65     Call to an undefined method Solarium\Component\Facet\FacetInterface::getField().
         🪪  method.notFound
  98     Call to an undefined method Solarium\Component\Facet\FacetInterface::serialize().
         🪪  method.notFound
 ------ -----------------------------------------------------------------------------------
```

PR #1186, opened 2026-07-24
Return type of field/function grouping value

thomascorthals:
Using `null` as an array key is deprecated in PHP 8.5. This is what happened when parsing a `Grouping` result if no grouping function is set. The function was always added to the array that's iterated and used as an array key inside the loop.

https://github.com/solariumphp/solarium/blob/ec841cf4134e1fb0399c9da63ed19acb426ce70b/src/Component/ResponseParser/Grouping.php#L46

I saw there was no example or integration test for grouping by function so I wanted to add those while I was at it. And that's when I noticed the group value was always returned as `?string` while functions can return `float` or `bool` (and grouping by a function returning `string` isn't even currently supported in Solr). The field value can also be `float`, `int`, or `bool` besides `string` or `null`. Coercing these to `string` seems odd when "Solarium is a PHP Solr client library that accurately models Solr concepts." I think it's more natural to keep the type that Solr returns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to the Solarium library will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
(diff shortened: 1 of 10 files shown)


mkalkbrenner:
nice, thank you

